PR #3, opened 2019-07-23
Packaging

joshbode:
Adds some packaging and an entrypoint: `datamodel-codegen`


koxudaxi:
@joshbode 
Thank you for the PR.
LGTM

joshbode:
Thank you! Looking forward to using this in a project at some point in the future :)

we're looking for a lightweight option for code-generating models for an API client for a private API and this looks like it could be very useful!

PR #69, opened 2019-10-10
Add missing imports for `constr` and `Dict[str, Any]`

joshbode:
also allows more than one import for `DataType` (required to support `Dict[str, Any]` for example)


koxudaxi:
Looks good 👍 
Thank you !!

PR #70, opened 2019-10-15
add support for object arrays

joshbode:
adds support for object as list element


koxudaxi:
It seems good :)

joshbode:
thank you!

koxudaxi:
The version 0.2.7 has been released.
Thank you very much!!

PR #71, opened 2019-10-15
Add support for custom templates and extra template data

joshbode:
- point to a directory containing template overrides
  (`BaseModel.jinja2`, etc)
- add custom data to templates for rendering from JSON file to support
  incorporating data that exists outside of openapi spec


koxudaxi:
I feel good.

koxudaxi:
Thank you very much.

joshbode:
Awesome - thank you again for accepting
 - sorry for the chain of PRs - I'm working up to something bigger (hint: it's an API client :)

koxudaxi:
@joshbode 

No problem :)
>  (hint: it's an API client :)

Will it be opened? (it's interesting)

I want to develop a code generator for a server of fastapi and client.
But, I'm busy with another project.

joshbode:
yes, I hope so :)

it's for an internal project at the moment (and as such is a little idiosyncratic, hence the desire to be able to modify templates a little, and inject some extra data) but as things settle down, I'll try to roll more of my steps back into the generator :)

koxudaxi:
If I will develop a code generator then, I will ping you. because we should save the resource to develop :)

PR #73, opened 2019-10-17
Fix some edge-cases with module references

joshbode:


koxudaxi:
Great work!!

koxudaxi:
It's released.
Thank you very much

PR #79, opened 2019-11-21
Support for Primitve union properties

joshbode:
For `anyOf` properties where the type is primitive (int, float, bool,
str, etc), produce a `Union` of primitive types without creating a
special "singular name" type


koxudaxi:
looks good :smile: 
Sorry,I was late to merge it.
I have released it as version 0.2.13.
Thank you very much.

joshbode:
You're welcome! Thanks for the merge

PR #80, opened 2019-11-22
add support for pyproject.toml if present

joshbode:


datamodel_code_generator/format.py
```diff
@@ -12,17 +14,28 @@
 
 
 def format_code(code: str, python_version: PythonVersion) -> str:
+
+    code = apply_isort(code)
```
joshbode, line 18: reversing this order (isort > black rather than black > isort) yields better results when there are imports directly in the (custom) Jinja template

koxudaxi: OK, I feel good.

koxudaxi:
@joshbode 

Looks good.
Would you please add documents to explain the content of `toml` file on READ.md?

joshbode:
Will do - by the way, isort already respects any pyproject.toml tool settings :)

----- Original message -----

joshbode:
I've added some text in the README about `pyproject.toml` - I also adopted black's function to find the `pyproject.toml`

koxudaxi:
I have released the version as 0.2.14.
Thank you very much.

PR #84, opened 2019-12-12
[FIX] types requiring imports in alloff, conint, confloat

krezac:
Hi Koudai,
thanks for yopur effort on that project, it's useful for us.
Here are couple of fixes I had to implement to be able to process our swagger file.
- if fields in unnamed object combined with ref in allOf require imports (i.e. string/date), these imports were not included. I'm not sure I've added it the best way, feel free to refactor. AnyOf seems to be fine, I've added couple of cases to tests.
- pydantic was not included for conint and confloat.

Regards
Kamil


koxudaxi:
@krezac 
I did not notice the case :( 

It looks good.
I have merged into master and I have released it as version  `0.2.16`

Thank you very much.

PR #91, opened 2020-02-03
Move logic for anyOf/allOf to higher priority

joshbode:
- `is_object` is masking `anyOf`/`allOf` logic preventing creation of `Union` fields


koxudaxi:
I had released a new version as 0.3.1. 
Thank you very much.

joshbode:
Great! Thanks @koxudaxi :)
I didn't dig deeply enough to figure out exactly what changed, but this worked :)

PR #92, opened 2020-02-04
add support for arrays of simple primities to anyOf

joshbode:
- map `{"type": "array", "items": {"type": primitive_type}}` to `List[py_primitive_type]`
  in `anyOf`


koxudaxi:
LGTM

koxudaxi:
I had already released the version as 0.3.2.
Thank you very much.

joshbode:
Great! Thank you for that - I think there's probably a more general way to do this than what I did here, but it's working for now :)

PR #94, opened 2020-02-26
Change numeric constraints to correct mapping

surajbarkale:
This change aligns mapping from json schema to pydantic Field arguments.
See https://pydantic-docs.helpmanual.io/usage/schema/#field-customisation


koxudaxi:
@surajbarkale 
I checked your code and the document. It's great.
I have released a new version as 0.3.3.

Thank you very much.

PR #98, opened 2020-03-09
correct string-normalization to skip-string-normalization

joshbode:
Fixing `skip-string-normalization` option to be in the right direction (negative rather than positive) to match the _actual_ configuration item name in Black
(sorry - the config file option is the opposite to the library option)


koxudaxi:
@joshbode 
I have released new version include the PR as `0.4.0`
Thank you very much

PR #110, opened 2020-04-01
handle edge-case where models map to the __init__.py level of a module

joshbode:
Sorry to follow up with a fix so soon after the previous PR - but as the branch name sez: `modules-are-hard` :)


joshbode:
Thank you!

koxudaxi:
I have released it as `0.4.4`!!

koxudaxi:
I know it is difficult to treat modular.
Thank you very much.

joshbode:
I keep finding new edge-cases as my use-cases evolve :)

PR #120, opened 2020-04-22
Fixed problems, if a name starts with a underscore

julian-r:
Names with underscore did not work, so here a small patch with a test


koxudaxi:
@julian-r 

Looks good!!

I have merged and released a new version as 0.4.9.

Thank you very much.

Issue #124, opened 2020-05-05
`black.InvalidInput: Cannot parse` error when pattern is explicit in a string property.


mgonzalezperna:
Using the JSONSchema way to declare a pattern for a string property fails on datamodel conversion.

-------

Based on the [JSONSchema docs](https://json-schema.org/understanding-json-schema/reference/regular_expressions.html#regular-expressions), the right way to declare a regex to validate the property is:
```
{
"rasen":
   "type": "string",
   "pattern": "^(\\([0-9]{3}\\))?[0-9]{3}-[0-9]{4}$"
}
```  
But using this in a property make datamodel-code-generator crash with a InvalidInput error because it cannot parse the arguments to use in the generated Field:

```
  File "/usr/lib/python3.8/site-packages/black.py", line 725, in format_str
    src_node = lib2to3_parse(src_contents.lstrip(), mode.target_versions)
  File "/usr/lib/python3.8/site-packages/black.py", line 836, in lib2to3_parse
    raise exc from None
black.InvalidInput: Cannot parse: 22:33:     rasen: Optional[constr(regex=^(\([0-9]{3}\))?[0-9]{3}-[0-9]{4}$)] = None
```

---------

I tried to do the same with a yml file and found that it crash in the same point.

```
ransen_property:
        type: string
        description: Test
        pattern: "[\/].*$"
```

The traceback in this case was:

```
  File "/usr/lib/python3.8/site-packages/black.py", line 725, in format_str
    src_node = lib2to3_parse(src_contents.lstrip(), mode.target_versions)
  File "/usr/lib/python3.8/site-packages/black.py", line 836, in lib2to3_parse
    raise exc from None
black.InvalidInput: Cannot parse: 35:28:     __root__: constr(regex=[/].*$)
```

I tracked the issue back to the type validator for strings and found that striping the quotes from the regex fires this error. I will make a PR with a suggested workaround to fix this.

PR #125, opened 2020-05-05
Fix parse errors when property pattern is explicit.

mgonzalezperna:
Proposed workaround to fix parse error when declaring regex. (https://github.com/koxudaxi/datamodel-code-generator/issues/124)

### DIFF
--- a/datamodel_code_generator/model/pydantic/types.py
+++ b/datamodel_code_generator/model/pydantic/types.py
@@ -134,6 +134,8 @@ def get_data_float_type(types: Types, **kwargs: Any) -> DataType:
 def get_data_str_type(types: Types, **kwargs: Any) -> DataType:
     data_type_kwargs = transform_kwargs(kwargs, string_kwargs)
     if data_type_kwargs:
+        if 'regex' in data_type_kwargs.keys():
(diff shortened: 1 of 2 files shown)


datamodel_code_generator/model/pydantic/types.py
```diff
@@ -134,6 +134,8 @@ def get_data_float_type(types: Types, **kwargs: Any) -> DataType:
 def get_data_str_type(types: Types, **kwargs: Any) -> DataType:
     data_type_kwargs = transform_kwargs(kwargs, string_kwargs)
     if data_type_kwargs:
+        if 'regex' in data_type_kwargs.keys():
```
koxudaxi: I think `.keys()` is not need.

koxudaxi:
@mgonzalezperna 
Thank you for creating This PR.

I wrote a comment, Otherwise, it's OK :smile:

datamodel_code_generator/model/pydantic/types.py
mgonzalezperna: Fixed!

koxudaxi:
@mgonzalezperna 
I have merged and published a new version as `0.4.10` :tada: 

Thank you very much :+1:

PR #138, opened 2020-06-10
the default for string-normalization was reversed

FlorianLudwig:
The default value for `skip-string-normalization` is `False` in black while datamodel-code-generator has it as `True`. To avoid confusion, use the same default value. Otherwise it is required to have a black config present with `skip-string-normalization` set to its default value.

### DIFF
--- a/datamodel_code_generator/format.py
+++ b/datamodel_code_generator/format.py
@@ -43,7 +43,7 @@ def apply_black(code: str, python_version: PythonVersion) -> str:
         mode=black.FileMode(
             target_versions={BLACK_PYTHON_VERSION[python_version]},
             line_length=config.get("line-length", black.DEFAULT_LINE_LENGTH),
-            string_normalization=not config.get("skip-string-normalization", True),
+            string_normalization=not config.get("skip-string-normalization", False),
         ),
     )
 


koxudaxi:
@FlorianLudwig 
Thank you for your great suggestion.
I agree we should set the same default.

However, This change is a large impact.
I think If a lot of users accept it, then we must change the option.
Also, We must change the major(minor) version when we merge the change.

I will create an issue to survey and discuss this problem.

FlorianLudwig:
I agree, this is indeed a breaking change as it might break the CI for others as well. Same goes for removing the `= None`. 


Side node:

How to properly define `skip-string-normalization`?

Using:
```
[tool.black]
skip-string-normalization = False
```

breaks `datamodel-code-generator`:
```
toml.decoder.TomlDecodeError: This float doesn't have a leading digit (line 2 column 1 char 13)
```

but using:
```
[tool.black]
skip-string-normalization = 0
```

breaks `black` cli:
```
  File "/home/f.ludwig/.local/share/virtualenvs/pdk8s-oowwi4Ru/lib/python3.8/site-packages/click/types.py", line 412, in convert
    value = value.lower()
AttributeError: 'int' object has no attribute 'lower'
```
([fixed in master, not yet released](https://github.com/psf/black/issues/1458))

koxudaxi:
@FlorianLudwig 

> (fixed in master, not yet released)
Sorry, I don't know the bug.

We should choose some options.
1. wait for the next release.
2. add a command-line option for string-normalization
3. add a workaround (example. Environment)

`1.` is clear. But, you may need the right now.
`2.` may be better than other options.
`3.` helps us for the time before the next release of black.

Would you think about it?

koxudaxi:
If some need to discuss,  then please creates a new PR.

PR #140, opened 2020-06-10
more appropriate bug_report template

FlorianLudwig:
update for the bug report template


koxudaxi:
@FlorianLudwig
Thank you very much :smile:

PR #141, opened 2020-06-10
dynamic field serialization

FlorianLudwig:
Make `DataModelField` generate it's string value dynamically instead of at creation time.

## Background
I have a few special cases and I want to avoid forking `datamodel-code-generator`, so my approach is currently like this: I have my own parser subclassing `JsonSchemaParser`:

```python
class K8SParser(JsonSchemaParser):
    def parse_raw(self):
        super(K8SParser, self).parse_raw()

        # manipulate self.results
```

I understand that the project is not meant to be used like this but hope my change is welcome anyway as it makes my use case easier :)

### DIFF
--- a/datamodel_code_generator/model/base.py
+++ b/datamodel_code_generator/model/base.py
@@ -53,7 +53,11 @@ class DataModelFieldBase(BaseModel):
     imports: List[Import] = []
     type_hint: Optional[str] = None
     unresolved_types: List[str] = []
-    field: Optional[str]
+
+    @property
+    def field(self) -> Optional[str]:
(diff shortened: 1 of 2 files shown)


koxudaxi:
@FlorianLudwig 
Thank you for creating the PR for the community.
The PR looks good.
Would you please run `./sciprts/format.sh` ?
Otherwise, OK.

FlorianLudwig:
@koxudaxi Ooops.

Done!

koxudaxi:
@FlorianLudwig 
I have released the version as `0.5.3`
Thank you very much.

Issue #139, opened 2020-06-10
`oneOf` not working inside `properties`


FlorianLudwig:
**Describe the bug**
I think it is allowed to use `oneOf` inside `properties` but it won't work.

**To Reproduce**

small.json
```json
{
    "definitions": {
      "io.k8s.api.admissionregistration.v1.MutatingWebhook": {
        "properties": {
          "timeout": {
            "oneOf": [
                {
                  "type": "string"
                },
                {
                  "type": "integer"
                }
            ]
          }
        },  
        "type": "object"
      }
    }
}
```

```bash
$ datamodel-codegen --input small.json --input-file-type jsonschema 
Traceback (most recent call last):
  [...]
  File "datamodel_code_generator/parser/jsonschema.py", line 158, in get_data_type
    raise ValueError(f"invalid schema object {obj}")
ValueError: invalid schema object items=None uniqueItem=None type=None format=None pattern=None minLength=None maxLength=None minimum=None maximum=None multipleOf=None exclusiveMaximum=None exclusiveMinimum=None additionalProperties=None anyOf=[] allOf=[] enum=[] writeOnly=None properties=None required=[] ref=None nullable=False x_enum_varnames=[] description=None title=None example=None examples=None default=None
```

koxudaxi:
@FlorianLudwig

Thank you for your report.
Yes, the feature is not implemented yet.
I think the generator should create validation method for `oneOf`.
I will implement it. I hope to try this week

koxudaxi:
I think we should resolve the problem(https://github.com/koxudaxi/datamodel-code-generator/issues/142) for this issue at first.

koxudaxi:
I  have released a new version as `0.5.6`

PR #147, opened 2020-06-11
support one_of

koxudaxi:
This PR support `oneOf`

## Related issue
https://github.com/koxudaxi/datamodel-code-generator/issues/139


tests/parser/test_jsonschema.py
```diff
@@ -142,6 +148,62 @@ def test_parse_object(source_obj, generated_classes):
...
+        for field in fields:  # type: ModelField
+            value, errors = field.validate(value, values={}, loc=(), cls=None)
+            if not errors:
+                if valid_item:
+                    raise ValueError(f'oneOf accept only one type. {value}')
+                valid_item = value
+                match = True
+        if match:
+            return valid_item
+        raise ValueError(f'oneOf accept only one type. {value}')
```
koxudaxi: @FlorianLudwig 
How do you think this validation logic for `oneOf`?

tests/parser/test_jsonschema.py
```diff
@@ -142,6 +148,62 @@ def test_parse_object(source_obj, generated_classes):
...
+                            }
+                        },
+                        "type": "object",
+                    }
+                }
+            },
+            """class Item(BaseModel):
+    timeout: Optional[Union[str, int]] = None
+
+    @validator('timeout', pre=True)
```
koxudaxi: I fixed it

tests/parser/test_jsonschema.py
FlorianLudwig: I couldn't get it working but I think the logic wont work.  If both validators are run after each other and it is an error if both are okay this will lead to false positives.  As pydantic as a parsing and not a validating library it does accept "1" as an integer. But also as an string. So it will raise an error.

Just having `Union[int, str]` without an extra validator and letting pydantic do its thing works fine for my use case.

koxudaxi: OK, I remove the validator. `Union` may not cover all use-case. but, It's good start point.

PR #159, opened 2020-06-20
clenaup openapi_parser_parse_alias test

FlorianLudwig:
cleaning up the test for better readability, especially on test failure.

### DIFF
--- a/tests/parser/test_openapi.py
+++ b/tests/parser/test_openapi.py
@@ -952,205 +952,219 @@ def test_openapi_parser_parse_alias():
         data_model_field_type=DataModelField,
         text=Path(DATA_PATH / 'alias.yaml').read_text(),
     )
-    assert parser.parse() == {
-        ('wo_o', '__init__.py'): '',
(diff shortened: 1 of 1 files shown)


koxudaxi:
@FlorianLudwig 

Thank you very much.

We may have to refactor all tests for better readability.
It may be an answer to write expected data in files.

How do you think?
example:
https://github.com/koxudaxi/fastapi-code-generator/blob/983c0bcb8978969a591bab242fbb6bcfa3c8e039/tests/test_generate.py

FlorianLudwig:
I like the idea of write expected data in files!

It might be a good chance to cleanup the tests overall - there seem to be quite a few tests testing the same thing.

One obvious example would be the `test_main` and `test_main_kr`. I think the `test_main` can be deleted as all tests are in the `test_main_kr` as well.

koxudaxi:
Thank you
I have created a new issue.

https://github.com/koxudaxi/datamodel-code-generator/issues/160

Issue #174, opened 2020-07-17
Handle double __ in snake to camel function


skrawcz:
# Problem
https://github.com/koxudaxi/datamodel-code-generator/blob/7eee45b5ba3db77c56bb56cae24f1ab3f7c88076/datamodel_code_generator/parser/base.py#L39-L45

It is an assumption that there is only a single `_` in field names. This can happen if there are two spaces in a name (https://github.com/koxudaxi/datamodel-code-generator/blob/master/datamodel_code_generator/parser/base.py#L142), or the field itself contains two `__`. 

So I think making an assumption that anything with `_` is valid snake case is incorrect. 

# Proposed solution
Change https://github.com/koxudaxi/datamodel-code-generator/blob/master/datamodel_code_generator/parser/base.py#L45 to 
```python
return prefix + ''.join(x[0].upper() + x[1:] for x in word.split('_') if x)
```
That way if there are multiple `_` they're skipped and the camelcase is still correctly created.

skrawcz:
Thanks!

PR #175, opened 2020-07-17
Implements #174

skrawcz:
Thanks for putting this library together.  Here's the implementation I was proposing in #174 to make the function more robust.


koxudaxi:
@skrawcz 
Thank you for creating the issue and this PR.
The suggestion is very good :+1: 

Would you please add a unittest for this case?

koxudaxi:
@skrawcz 
I have merged and released it as version `0.5.15`. :tada: 

Thank you very much :smile:

Issue #182, opened 2020-07-27
yaml support is not complete


ioggstream:
**Describe the bug**
Parsing yaml OAS3 files does not work

**To Reproduce**
See #183 .

Example schema:
```yaml
components:
  schemas:
    Problem:
      $ref: https://teamdigitale.github.io/openapi/0.0.6/definitions.yaml#/schemas/Problem

```

Used commandline:
```
$ datamodel-codegen test.yaml
``

**Expected behavior**
A generated Problem class

**Version:**
 - OS: linux
 - Python Verison: 3.7
 - datamodel-code-generator Version latest

PR #183, opened 2020-07-27
Fix: #182. Fix yaml support. Improve OAS3

ioggstream:
## This PR

Fix yaml support
Improves OAS3
Use snaphots for tests


datamodel_code_generator/parser/jsonschema.py
```diff
@@ -633,10 +632,21 @@ def parse_raw_obj(self, name: str, raw: Dict[str, Any], path: List[str]) -> None
         self.parse_ref(obj, path)
 
     def parse_raw(self) -> None:
-        raw_obj: Dict[Any, Any] = json.loads(self.text)  # type: ignore
```
ioggstream, line 636: Always use yaml parser so we don't care about file format.

datamodel_code_generator/parser/jsonschema.py
```diff
@@ -633,10 +632,21 @@ def parse_raw_obj(self, name: str, raw: Dict[str, Any], path: List[str]) -> None
...
         obj_name = self.model_resolver.add([], obj_name, unique=False).name
         self.parse_raw_obj(obj_name, raw_obj, [])
-        definitions = raw_obj.get('definitions', {})
+        definitions, path = get_oas_schema(raw_obj)
         for key, model in definitions.items():
-            self.parse_raw_obj(key, model, ['#/definitions', key])
+            self.parse_raw_obj(key, model, [path, key])
+
+
+def get_oas_schema(oas_spec: Dict[Any, Any]):
```
ioggstream: Support for different OAS schemas:
OAS2 - #/definitions
OAS3- #/components/schemas
...

datamodel_code_generator/parser/jsonschema.py
koxudaxi: Great :+1:

datamodel_code_generator/parser/jsonschema.py
koxudaxi: Sorry, I forgot the difference.
This file is for `JSON Schema`.
this method name should not have `OSA`.
Would you please remove `OSA` from the method ?

koxudaxi:
@ioggstream 
I reviewed it.
Otherwise ok :smiley_cat:

datamodel_code_generator/parser/jsonschema.py
ioggstream: I just noticed you had another class for OAS3, and I moved out stuff there.
Should be fixed now.
We'll find a way to fix the usage of `definitions` in jsonschema another time :)

ioggstream:
Should be ok. Take some time to review in the next days. :D

datamodel_code_generator/parser/jsonschema.py
koxudaxi: This code generator supported only OSA3 for the first time.
But, I supported JSON Schema by a request.

We will fix it next time:wink:

koxudaxi:
I think this PR is fine.

I have merged this PR because I prefer a short iteration in this phase.

koxudaxi:
@ioggstream 
I have released the version as `0.5.20`

Thank you very much.

PR #194, opened 2020-08-06
Add uri-reference

drsm79:
Added as a string, something more path like via a `ConstrainedStr` might be an improvement in the future. Let me know if I should add some tests somewhere, I didn't see any for the `uri`.

### DIFF
--- a/datamodel_code_generator/parser/jsonschema.py
+++ b/datamodel_code_generator/parser/jsonschema.py
@@ -59,6 +59,7 @@ json_schema_data_formats: Dict[str, Dict[str, Types]] = {
         'uuid4': Types.uuid4,
         'uuid5': Types.uuid5,
         'uri': Types.uri,
+        'uri-reference': Types.string,
         'hostname': Types.hostname,
         'ipv4': Types.ipv4,
         'ipv6': Types.ipv6,
--- a/tests/parser/test_openapi.py
+++ b/tests/parser/test_openapi.py
@@ -64,6 +64,7 @@ class A(TemplateBase):
         ('string', 'password', 'SecretStr', 'pydantic', 'SecretStr'),
         ('string', 'email', 'EmailStr', 'pydantic', 'EmailStr'),
         ('string', 'uri', 'AnyUrl', 'pydantic', 'AnyUrl'),
+        ('string', 'uri-reference', 'str', None, None),
         ('string', 'uuid', 'UUID', 'uuid', 'UUID'),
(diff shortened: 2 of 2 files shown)


koxudaxi:
@drsm79 
 I'm sorry for the late reply.

There are unit tests for data types.
https://github.com/koxudaxi/datamodel-code-generator/blob/1b8ae11db16864ab26d58030f1cd447111c09a79/tests/parser/test_openapi.py#L77

koxudaxi:
@drsm79 

> Is there a path-like type it could be, else a string makes sense.

I answer it here.

There is [Path](https://docs.python.org/3/library/pathlib.html) in Python standard library.
And You can use `Union` to support multiple types.

```python
from pathlib import Path
from typing import Union

from pydantic import AnyUrl, BaseModel


class Model(BaseModel):
    ref_uri: Union[AnyUrl, Path]


print(Model(ref_uri='https://example.com'))
# ref_uri=AnyUrl('https://example.com', scheme='https', host='example.com', tld='com', host_type='domain')

print(Model(ref_uri='xyz/abc.csv'))
# ref_uri=PosixPath('xyz/abc.csv')
```

But,  `get_data_type` does not support the `Union` type.
https://github.com/koxudaxi/datamodel-code-generator/blob/1b8ae11db16864ab26d58030f1cd447111c09a79/datamodel_code_generator/model/pydantic/types.py#L174

I think this PR must be minimum implementation (only `str` or `AnyUrl`).
We will be able to support more improvement types like `Union[AnyUrl, Path]`  in next chance.

drsm79:
Thinking about it a Path wouldnt work if the uri-reference was a full uri. I'll make it a string :)

koxudaxi:
@drsm79 
I think It's a good choice. :+1: 

I have released a new version as 0.5.25.
Thank you very much.


Also, I just came up with a fallback type.
If this generator doesn't know a format then, the generator uses the default type as a fallback type.
How do you think about it?

drsm79:
Cool, thanks :) 

Defaults are always a bit awkward, especially in a schema based thing. Maybe something a user can provide?

koxudaxi:
Sorry, `default` is not a good word. 

I expected this pattern.
If I use an unknown format for string type then, the code generator drops format.  
example schema:
 `{"type": "string", format: "user-custom-format" }` -> `{"type": "string"}`

But, We should notify the behavior to the user via stdout.
example message:
```
the unknown format was parsed as a string type (format:  "user-custom-format")
```

However, As you said the default is awkward on schema-based things.
users may want to give the default as CLI options :thinking:

PR #199, opened 2020-08-16
Added support of integer in string

Forden:
This covers this case:

{"some_data": "123"}


koxudaxi:
@Forden 

I have released a new version as 0.5.27 :tada: 

Thank you very much.

PR #203, opened 2020-08-27
Fix nested refs and subclasses

wadedyck:
This fixes model generation for openapi specs with nested references or with parent classes different than the base_model.

Attached is the Jetbrain's youtrack openapi.json that I used.

[openapi.zip](https://github.com/koxudaxi/datamodel-code-generator/files/5138984/openapi.zip)

### DIFF
--- a/datamodel_code_generator/parser/base.py
+++ b/datamodel_code_generator/parser/base.py
@@ -88,10 +88,17 @@ def sort_data_models(
     if require_update_action_models is None:
         require_update_action_models = []
 
+    deferred_children: List[DataModel] = []
     unresolved_references: List[DataModel] = []
     for model in unsorted_data_models:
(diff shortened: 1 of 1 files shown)


koxudaxi:
@wadedyck 
I think good.

Would you please add a unittest?
Also, We should make the coverage 100%

Sorry, Travis-ci doesn't notify the field status on this PR.  
https://travis-ci.org/github/koxudaxi/datamodel-code-generator/builds/721842629

koxudaxi:
@wadedyck 
Would can you add unit tests on this PR?
If you are busy then, I will add it when I can get the time.

koxudaxi:
I have released a new version as `0.6.13` 
I fixed the problem on another PR.

Thank you very much

Issue #223, opened 2020-09-21
KeyError: 'array' when null values are interspersed with [] for the same dictionary key


JosXa:
**Describe the bug**

When trying to parse a JSON list of objects where items have both `null` and a list of values (or an empty list) for the same key, the library raises a `KeyError` (see example which should make it easier).

Something like this was supposedly fixed in response to #164, so it's either a regression or a new bug.

**To Reproduce**

Example JSON:

```json
{
  "my_obj": [
    {
      "items": null
    },
    {
      "items": []
    }
  ]
}
```

It doesn't matter whether the null or the [] is defined first.

The schema generated from this is:

```json
{
    "$schema": "http://json-schema.org/schema#",
    "type": "object",
    "properties": {
        "my_obj": {
            "type": "array",
            "items": {
                "type": "object",
                "properties": {
                    "items": {
                        "type": [
                            "array",
                            "null"
                        ]
                    }
                },
                "required": [
                    "items"
                ]
            }
        }
    },
    "required": [
        "my_obj"
    ]
}
```

Used `generate` call:
```
            generate(
                input_name=file.name,
                input_text=content,
                input_file_type=InputFileType.Json,
                output=out_file,
                target_python_version=PythonVersion.PY_38,
                snake_case_field=True,
            )
```

Traceback:

```
Traceback (most recent call last):

  File "C:\projects\josxabot\app\modules\josxa_private\pydantic_model_gen\directory_watcher.py", line 92, in run_conversion
    generate(
    └ <function generate at 0x00000251AF436CA0>

  File "C:\Users\INT002327\AppData\Local\pypoetry\Cache\virtualenvs\josxabot-c3BmTbt9-py3.8\lib\site-packages\datamodel_code_generator\__init__.py", line 185, in generate
    result = parser.parse()
             │      └ <function Parser.parse at 0x00000251AF436A60>
             └ <datamodel_code_generator.parser.jsonschema.JsonSchemaParser object at 0x00000251AF20E6A0>

  File "C:\Users\INT002327\AppData\Local\pypoetry\Cache\virtualenvs\josxabot-c3BmTbt9-py3.8\lib\site-packages\datamodel_code_generator\parser\base.py", line 351, in parse
    self.parse_raw()
    │    └ <function JsonSchemaParser.parse_raw at 0x00000251B1703430>
    └ <datamodel_code_generator.parser.jsonschema.JsonSchemaParser object at 0x00000251AF20E6A0>

[...]

PR #226, opened 2020-09-23
support null and array in type

koxudaxi:
This PR supports null and array in type
## Related Issues
https://github.com/koxudaxi/datamodel-code-generator/issues/223


koxudaxi:
@JosXa
Would you please review this PR?

datamodel_code_generator/parser/jsonschema.py
```diff
@@ -183,10 +184,21 @@ def get_data_type(self, obj: JsonSchemaObject) -> List[DataType]:
         if isinstance(obj.type, list):
             types: List[str] = [t for t in obj.type if t != 'null']
```
JosXa, line 185: I'm curious if the string `'null'` needs to be stripped here for some weird inputs (e.g. `'  nUlL     '`)

JosXa (approved):
Cool, thanks a ton!
I can't pretend I understand a lot of what's happening here, but the test case looks good!

datamodel_code_generator/parser/jsonschema.py
koxudaxi: I guess I don't need stripping these inputs.
`null` is defined in the rule book.

koxudaxi:
@JosXa 
Thank you for reviewing the PR.
> I can't pretend I understand a lot of what's happening here, but the test case looks good!
Yes, This parser is a little complex structure to support a few formats in the future.

Of course,  We can refactor the parser.

JosXa:
Might be time.. One thing I noticed is that the code tends to repeat the same name for local vars and self vars and then even other objects with the same names. Think:

```
...
my_var = [x for x in self._my_var]
some_obj.my_var = my_var
```
This made it hard to read. Food for thought :)

koxudaxi:
> Might be time.. One thing I noticed is that the code tends to repeat the same name for local vars and self vars and then even other objects with the same names. Think:

It's my coding style recently...
I agree with your thought. I should change it to read easily. :smiley_cat: 
I want to change all the similar code when I get the chance. 

If you don't have other pointed out then I merge it.

JosXa:
Cool hehe, and glad to be working with a friendly maintainer :) Let's see if I can give this a shot with PRs next time I find a bug 💪😁

LGTM!

koxudaxi:
I'm happy working with you too :smile: 
PRs welcome :rocket:

Issue #195, opened 2020-08-13
Change install requirements from pinpoint to lower bounds


dennis-wey:
Hi,

first of all thank you for the project this looks really interesting!

**Is your feature request related to a problem? Please describe.**
All the versions in requirements are pinpoint which might cause version conflicts when using this with other packages e.g. fastapi. 

**Describe the solution you'd like**
Instead of pinpoint the pydantic version make it lower bound
`pydantic[email]>=1.5.1`

koxudaxi:
@dennis-wey 
Thank you for creating this issue.

I'm afraid of broken dependency in the future.

> All the versions in requirements are pinpoint which might cause version conflicts when using this with other packages e.g. fastapi.

How did you use this code-generator? as a module? or CLI?

If you use it as CLI then, I recommend that you should not mix other packages in the environment.
I usually create a requirements.txt(or poetry or pipenv) for only this code generator.
Also, I may choice [pipx](https://github.com/pipxproject/pipx) for the standalone app.

However, pinpoint versions are not comfortable for us.
We should set a cap version when we cut the pinpoint version  (example: "pydantic>=1.51,<2.0")

How do you think about it?

dennis-wey:
My use case is, that for a bigger project I would like to build pydantic classes out of openapi.json provided by a fastapi application. I don't want to automate the creation of the whole client but make the schemas available. So my use case would be using a parser and then accessing its `results` attribute. So basically I wouldn't even need black or isort package but since their vital to the core use case, it wouldn't make sense to make them optional.

> We should set a cap version when we cut the pinpoint version (example: "pydantic>=1.51,<2.0")
Yes this sounds like a very good approach.

koxudaxi:
@dennis-wey 

> So my use case would be using a parser and then accessing its results attribute. 

I'm happy you use this library deeply :smile: 

> Yes this sounds like a very good approach.

Thank you
I have changed and released a new version as `0.5.26`  :rocket:

dennis-wey:
Sorry me again. I did tried to include the project in the bigger project but got stuck again since due to strict requirement managament I can only rely on argcomplete ~=1.10. Is it reasonable to lower the minimal version on this one?

Also for me it seems like this package does two things:

1. Providing Parser forbuilding pydantic Classes out of openapi/json schema.
2. Provide a code generator command line tool

[...]

PR #251, opened 2020-11-08
update isort and black

Chilipp:
Hey @koxudaxi!

very nice package, thanks a lot for your work on this! I'd like to use it and I am already using the latest version of isort and black. It would be nice if the datamodel-code-generator is compatible with this which is whiy I make this PR.

From my perspective, this PR can be merged if the CI passes.


Chilipp:
hmm ok. the CI is failing because a new version of `black` is installed. But to me, this is actually rather a CI issue and does not justify to force the users of this package to install an old version of `black`.

not sure what would be the best to fix this. One option would be to create a venv in [`lint.sh`](https://github.com/koxudaxi/datamodel-code-generator/blob/master/scripts/lint.sh) to make sure that the correct version of black and isort is installed

```bash
#!/usr/bin/env bash

set -e

python -m venv lint
source lint/bin/activate

pip install black==19.10b0 "isort>=4.3.21,<5.0"

black --check datamodel_code_generator tests
isort --check-only datamodel_code_generator tests

deactivate
rm -r lint

mypy datamodel_code_generator

```

what do you think @koxudaxi?

koxudaxi:
@Chilipp 
Thank you for creating this issue.
I understand your suggestion.
But, some users want to keep the old version of dependence packages.
we talked about the problem.
https://github.com/koxudaxi/datamodel-code-generator/issues/195
Can you support the old version too?

I have checked your PR.
A related code of isort is changed.
I would we can check the version before run sort.
```python
if isort.__version__.startswith('4.'): 
   return SortImports(file_contents=code).output
else:
   return isort.code(code) 
```

> not sure what would be the best to fix this. One option would be to create a venv in lint.sh to make sure that the correct version of black and isort is installed

I think the change is good :+1:

Chilipp:
> I would we can check the version before run sort.

alright, done with https://github.com/koxudaxi/datamodel-code-generator/pull/251/commits/cc25f6966811c4653b35801d2dc22829bf4ac942

> I think the change is good +1

ok, see https://github.com/koxudaxi/datamodel-code-generator/pull/251/commits/6a2397cee7b3c507f70ded066b9c0b00e8cf8edd

Chilipp:
alright, the same problem now with `test.sh` and `test.bat`. I'd fix this with something like

```bash
#!/usr/bin/env bash
set -e

python -m venv venv

source venv/bin/activate
pip install -e .[all] isort==4.3.21 "black>=19.10b0,<20"

pytest --cov=datamodel_code_generator --cov-report term-missing tests

deactivate
rm -r venv
```

what do you think @koxudaxi ?

koxudaxi:
@Chilipp 
I think your approach is good :rocket:

Chilipp:
alright, seems to work @koxudaxi :smile: I am done here, merge it if you like

Thanks a lot again for your work on this =)

koxudaxi:
I just came up with a good approach.
If `list.sh` have to change arguments of isort by versions, we can check the isort version in the shell scripts and use the `if` statement. 

I may change the shell scripts.

Thank you very much.

Chilipp:
sorry, I can't really follow, but do what you think is the best @koxudaxi :sweat_smile:

koxudaxi:
No problem. I was too late getting the idea :sweat_smile: 
I will release tonight:rocket:

Chilipp:
awesome! thanks a lot =)

koxudaxi:
@Chilipp 
I have released a new version as `0.6.3`

PR #254, opened 2020-11-09
fix settings for isort

koxudaxi:


koxudaxi:
@Chilipp 
I meet a problem with isort 5.x
The result of the unittest is different from version 4.x.

version 4.0
```
from pydantic import Extra

from custom_module import Base
```

version 5.0
```
from custom_module import Base
from pydantic import Extra
```

I guess isort don't detect pydantic as third-party modules.

I will think about how do I resolve the problem.

Chilipp:
hmm I see @koxudaxi. You likely have to add it as a known third-party module (https://pycqa.github.io/isort/docs/configuration/options/#known-third-party)

koxudaxi:
@Chilipp 
I just fixed it.
Thank you!!

PR #273, opened 2020-12-03
Warn with default format for type when not found

Liam-Deacon:
This is a workaround for an error I encountered whilst processing OpenAPI YAML schemas where the string format was a regular expression rather than one of the set of predefined jsonschema format types.

This PR issues a warning to the user that the format of the type is not understood and that the default format will be used instead, thus producing generated python files when previously only a traceback was shown.


koxudaxi:
@Liam-Deacon 
I feel great about this PR.
Would you please format the code to pass CI? 
the description is this link.
https://koxudaxi.github.io/datamodel-code-generator/development-contributing/

koxudaxi:
I improved your PR.
I have released a new version as 0.6.13

Thank you very much

PR #279, opened 2020-12-18
Fixed misreferenced examples in the docs.

blockjoe:
It looks like the bash examples were backwards for the with/without `--field-constraints`, I just swapped them.


koxudaxi:
@blockjoe
Great!! 
Thank you very much.

PR #408, opened 2021-04-05
First Round Automated API Model schema code Generator changes for Schematics

pleonowens:
This first round is functional, with some extra code in output.py for enums, and a few extra imports, but all the needed functionality is there. This will be put up for review with open source after it's cleaned up


pleonowens:
Meant to open this PR in a local fork, will re-open changes to here when its been finalized

PR #421, opened 2021-04-29
Fix #420

### LINKED ISSUE #420: IPv4Address doesn't import from pydantic.validators
**Describe the bug**

When using `format: ipv4`, the following import is added to the output:

```py
from pydantic import IPv4Address
```

This isn't a valid import.

**To Reproduce**

Example schema:
```yaml
openapi: 3.0.0

info:
  version: 0.0.1
  title: Foo API

paths:
  /foo:
    get:
      responses:
        "200":
          description: Success

components:
  schemas:
    Foo:
      type: object
      properties:
        ip:
          type: string
          format: ipv4
```

Used commandline:
```
$ datamodel-codegen --input openapi.yaml
```

**Expected behavior**

When using `format: ipv4`, the following import is added to the output:

```py
from pydantic.validators import IPv4Address
```

**Version:**
 - OS: MacOS
 - Python version: `3.9.2`
 - datamodel-code-generator version: `0.8.2`

**Additional context**
None

goodoldneon:
Make `IPv4Address` and `IPv6Address` import from `pydantic.validators`.


datamodel_code_generator/model/pydantic/imports.py
```diff
@@ -17,8 +17,8 @@
 IMPORT_UUID4 = Import.from_full_path('pydantic.UUID4')
 IMPORT_UUID5 = Import.from_full_path('pydantic.UUID5')
 IMPORT_ANYURL = Import.from_full_path('pydantic.AnyUrl')
-IMPORT_IPV4ADDRESS = Import.from_full_path('pydantic.IPv4Address')
-IMPORT_IPV6ADDRESS = Import.from_full_path('pydantic.IPv6Address')
+IMPORT_IPV4ADDRESS = Import.from_full_path('pydantic.validators.IPv4Address')
+IMPORT_IPV6ADDRESS = Import.from_full_path('pydantic.validators.IPv6Address')
```
koxudaxi: @goodoldneon 
In official documents ,It's written `ipaddress.IPv4Address` ,`ipaddress.IPv6Address`
https://pydantic-docs.helpmanual.io/usage/types/#pydantic-types
I have not used it.
Where should we import it from? 🤔

amh4r: Oh, it looks like it should be imported from the `ipaddress` lib. It isn't defined by Pydantic:

https://github.com/samuelcolvin/pydantic/blob/5261fd05a0374b84ce2602d45990baf480fa2417/pydantic/validators.py

That's probably why the docs say `ipaddress.IPv4Address`

koxudaxi: OK, Would you please change it?

koxudaxi:
@goodoldneon 
Also, We should fix the test.
https://github.com/koxudaxi/datamodel-code-generator/blob/d57b73895cf749118e93891e9bdf7070bede73fe/tests/parser/test_jsonschema.py#L357-L358

koxudaxi:
@goodoldneon 
I have released a new version `0.11.3`
Thank you very much 😄

PR #450, opened 2021-06-14
Update __init__.py

rgreinho:
Adds directives to prevent `flake8` and `pylint` from linting the generated file containing the models.


koxudaxi:
@rgreinho 
Thank you for creating this issue.
Would you please add the feature as a CLI option?

rgreinho:
I mean, yes, I can do that, but since the file will never pass the checks, unless you make very specific configuration adjustments, I am not sure I see the point of not having them there permanently. Can you clarify what the benefit would be?

koxudaxi:
@rgreinho 
The code-generator is used by many projects. It's downloaded about 100k per month.
I guess the generated files are managed by git.
they have to accept these lines for pylint and flake8.
The changes will increase the burden on maintainers of projects :cry: 
Also, the changes may break CI in some projects.

What did you think about it?

PR #521, opened 2021-08-29
Add playground link

dukkee:
Hello, @koxudaxi Thank you for your great tool. I think, it will be useful to give users an opportunity to try your generator online. Therefore I created a simple (for now) playground for your tool :smiley: http://dcg-playground.boardpack.org/ 

Future improvements you can find in the repo: https://github.com/boardpack/dcg-playground


koxudaxi:
@dukkee 
I'm sorry for my late reply.
Could you please tell me about the system?
Where is run the app in? How is deployed?

Thank you

dukkee:
Hello @koxudaxi What do you mean by the "tell about the system"? I sent you the repo link. Currently, the system uses a simple free Heroku instance

koxudaxi:
@dukkee 
OK, I see.
Thank you for providing the system.
I'm afraid two things
1. I can't fix and deploy the app. But, you write `with our online [playground]...`
We should indicate is not an official site.
2. The system doesn't show the version. This app will be released short term.
We should show the version on your site.

What do you think about it?

dukkee:
@koxudaxi Sorry for this late answer..

1. I can share repo and Heroku project, it's not a problem
2. It makes sense, you are entirely right, thanks for the tip.

koxudaxi:
@dukkee 

Thank you for the comments.

> 1. I can share repo and Heroku project, it's not a problem
> 2. It makes sense, you are entirely right, thanks for the tip.

Could you please add the information?

koxudaxi:
no response :(

Issue #593, opened 2021-10-31
Failing to parse open-rpc's meta-schema


vesajaaskelainen:
**Describe the bug**
open-rpc provides meta-schema in following locations:
https://github.com/open-rpc/meta-schema/
and 
https://meta.open-rpc.org/

Now in order to parse files based on this meta-schema one would like to generate data model for it.

There seems to be multiple errors in handling open-rpc meta-schema and referenced json-schema-tools's meta-schema:
- reference is not working with absolute https uri.
- json-schema-tools's schema.json 'items' keyword in 'enum' and in 'examples' array which does not work
- json-schema-tools's schema.json 'const' keyword does not work
- json-schema-tools's schema.json 'default' keyword in JSONSchemaObject does not work

**To Reproduce**

Example schema:
```shell
# use directly from web: 
# https://meta.open-rpc.org/
# or
# download locally:
wget https://meta.open-rpc.org/ -O open-rpc-meta-schema.json
# or fetch latest release from github:
wget https://github.com/open-rpc/meta-schema/releases/download/1.14.2/open-rpc-meta-schema.json
```

Used commandline:
```
(venv) user@host:/tmp/test$ datamodel-codegen --url https://meta.open-rpc.org/ --input-file-type jsonschema --output openrpc.py
Traceback (most recent call last):
  File "/tmp/test/venv/lib/python3.8/site-packages/datamodel_code_generator/__main__.py", line 495, in main
    generate(
  File "/tmp/test/venv/lib/python3.8/site-packages/datamodel_code_generator/__init__.py", line 353, in generate
    results = parser.parse()
  File "/tmp/test/venv/lib/python3.8/site-packages/datamodel_code_generator/parser/base.py", line 426, in parse
    self.parse_raw()
  File "/tmp/test/venv/lib/python3.8/site-packages/datamodel_code_generator/parser/jsonschema.py", line 1177, in parse_raw
    self._parse_file(self.raw_obj, obj_name, path_parts)
  File "/tmp/test/venv/lib/python3.8/site-packages/datamodel_code_generator/parser/jsonschema.py", line 1255, in _parse_file
    self.parse_raw_obj(key, model, path)
  File "/tmp/test/venv/lib/python3.8/site-packages/datamodel_code_generator/parser/jsonschema.py", line 1122, in parse_raw_obj
    self.parse_obj(name, JsonSchemaObject.parse_obj(raw), path)
[...]

PR #594, opened 2021-10-31
Fix parsing of absolute reference URLs

vesajaaskelainen:
If URL was referenced like:
{"$ref": "https://server/schema.json"}

It would result lookup by:
<original file/url>/https://server/schema.json

Change fixes so that new absolute URL will be used.

Relates to issue: https://github.com/koxudaxi/datamodel-code-generator/issues/593


koxudaxi:
@vesajaaskelainen  
I have released new version [0.11.15](https://github.com/koxudaxi/datamodel-code-generator/releases/tag/0.11.15)

Issue #508, opened 2021-08-20
pip install datamodel-code-generator[http] not working


eg-firebolt:
**To Reproduce**

In a fresh virtualenv, I am not able to install with [http] option as described by the README. 

```shell
❯ pip install datamodel-code-generator[http]
zsh: no matches found: datamodel-code-generator[http]
```

Usual install without http option works fine

**Version:**
 - OS: macOS
 - Python version: 3.9

koxudaxi:
@eg-firebolt 
Thank you for creating this issue.
You should escape the `[ ]` because zsh detects `[ ]` as a special character.

```
$ pip install datamodel-code-generator\[http\]
```
or 
```
$ pip install "datamodel-code-generator[http]"
```

We should write the topic on README.

PR #620, opened 2021-11-19
documented installation via pip

adaamz:
fixes https://github.com/koxudaxi/datamodel-code-generator/issues/508


koxudaxi:
@adaamz
Thank you for adding the document 🎉

PR #670, opened 2022-01-03
WIP: Differentiate "nullable" and "not-required" properties

goodoldneon:
> :warning: This PR isn't done yet. I'm creating it to get some feedback and help.

## Definition of Done

Add support for differentiating between "nullable" and "not-required" properties. For example, the following spec generates the following model:

```yaml
components:
  schemas:
    Pet:
      type: object
      required:
        - id
        - name
      properties:
        id:
          type: integer
          format: int64
        name:
          type: string
        tag:
          type: string
```

```py

from typing import Optional, Union

class Unset:
    pass


unset = Unset()


class Pet(BaseModel):
    id: int
    name: str
    tag: Union[str, Unset] = unset
```

## Missing Stuff

- CLI argument for opting into this new behavior.
- Adding the `typing.Union` import.

### DIFF
--- a/datamodel_code_generator/model/base.py
+++ b/datamodel_code_generator/model/base.py
@@ -28,6 +28,7 @@ from datamodel_code_generator.types import DataType, chain_as_tuple
 TEMPLATE_DIR: Path = Path(__file__).parents[0] / 'template'
 
 OPTIONAL: str = 'Optional'
+UNION: str = 'Union'
 
 ALL_MODEL: str = '#all#'
 
@@ -64,6 +65,8 @@ class DataModelFieldBase(_BaseModel):
 
         if not type_hint:
             return OPTIONAL
+        elif self.required is False:
+            return f'{UNION}[{type_hint}, Unset]'
         elif self.nullable is not None:
             if self.nullable:
                 return f'{OPTIONAL}[{type_hint}]'
--- a/datamodel_code_generator/model/template/pydantic/BaseModel.jinja2
+++ b/datamodel_code_generator/model/template/pydantic/BaseModel.jinja2
@@ -24,7 +24,9 @@ class {{ class_name }}({{ base_class }}):
     {%- else %}
     {{ field.name }}: {{ field.type_hint }}
     {%- endif %}
-    {%- if not (field.required or (field.represented_default == 'None' and field.strip_default_none))
+    {%- if not field.required
+            %} = unset
+    {%- elif not (field.required or (field.represented_default == 'None' and field.strip_default_none))
             %} = {{ field.represented_default }}
     {%- endif -%}
     {%- endif %}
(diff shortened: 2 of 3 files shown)


datamodel_code_generator/model/base.py
```diff
@@ -64,6 +65,8 @@ def type_hint(self) -> str:
 
         if not type_hint:
             return OPTIONAL
+        elif self.required is False:
+            return f'{UNION}[{type_hint}, Unset]'
```
koxudaxi, line 69: @goodoldneon 
Should we use a built-in special `unset` class in Python or Pydantic?
I'm sorry that I forget the name of the special `unset` class :(
Do you know the name?

koxudaxi:
I think The idea is a good start point.

datamodel_code_generator/model/base.py
amh4r: If there's a builtin class we should definitely use it! But I'm not able to find one 🙁

lafrenierejm:
No specific suggestions, but some references:
- Pydantic models have a [`__fields_set__` attribute](https://pydantic-docs.helpmanual.io/usage/models/#model-properties) that can be used to determine whether a given field was explicitly populated or is merely using the default.
- [This comment](https://github.com/samuelcolvin/pydantic/issues/1223#issuecomment-585371034) on [Pydantic #1223](https://github.com/samuelcolvin/pydantic/issues/1223) provides _a_ pattern for optional, non-nullable properties.
- Further in the [Pydantic #1223](https://github.com/samuelcolvin/pydantic/issues/1223) thread there are links to PRs that propose more complete solutions.

datamodel_code_generator/model/base.py
karolzlot: Maybe python's ellipsis would be good fit for this? https://stackoverflow.com/a/47189816/8896457

But it is used already by pydantic, I'm not sure if things won't collide.

karolzlot: Please take a look also at this discussion: 
https://chat.stackoverflow.com//transcript/message/54521781#54521781

karolzlot: And also this: https://peps.python.org/pep-0661/

karolzlot: Pydantic has its own `Undefined` type:

https://github.com/samuelcolvin/pydantic/blob/cc54acb612cb5144a34caebb9ac143324a0cb4a1/pydantic/fields.py#L81

It is now used also in FastAPI (today's update):

https://github.com/tiangolo/fastapi/blob/1876ebc77949a9a254909ec61ea0c09365169ec2/fastapi/dependencies/utils.py#L46

karolzlot:
@goodoldneon

I noticed that you closed your pull request. Can you let me know why you decided to close it and if you have any plans for it in the future?

Thanks for your help!

PR #679, opened 2022-01-10
Use `Non{Positive,Negative}{Float,Int}` in models

duesenfranz:
Expands the usage of `{Positive,Negative}{Float,Int}` to also support
`Non{Positive,Negative}{Float,Int}` in code generation.


koxudaxi:
@duesenfranz 
I have not heard the types :)

I think the changes affect current users.
Can you add the feature as a cli option?

duesenfranz:
I'll try! I guess the cli option should be off by default, right? :)

duesenfranz:
Also note that while generated files change, the changes should not break anything: these types are just the correct `ConstrainedFloat`/`ConstrainedInt` types - with the upside of not breaking typecheckers.

koxudaxi:
@duesenfranz 

> I'll try! I guess the cli option should be off by default, right? :)

Yes

> Also note that while generated files change, the changes should not break anything: these types are just the correct ConstrainedFloat/ConstrainedInt types - with the upside of not breaking typecheckers.

I explain why I would avoid changing default behavior.
I guess the CLI is used in many projects. If we change the default output then many projects have to change many lines. I expected the generated code is managed in git. Many people want to prevent unnessecey changes.

duesenfranz:
@koxudaxi
I added a command line argument `use_typecheck_friendly_constrained_types` which is off by default to use the feature.
I know the name is quite long; I can change it if you have another name in mind :)

I will squash the changes before merging!

koxudaxi:
@duesenfranz 
> I added a command line argument use_typecheck_friendly_constrained_types which is off by default to use the feature.
> I know the name is quite long; I can change it if you have another name in mind :)
Thank you for changing it.
OK, I will change the name 😉

koxudaxi:
I just changed the name to `use_non_positive_negative_number_constrained_types`

duesenfranz:
@koxudaxi thank you! :)

PR #689, opened 2022-01-20
do not convert literal int to string

### LINKED ISSUE #688: --enum-field-as-literal=one converts integer to string - still
**Describe the bug**
When using `--enum-field-as-literal=one`, literal integers get converted to strings, depending on the exact version of `datamodel`'s dependencies which are installed. For details see the bottom of the description. This is highly problematic when using `datamodel-code-generator` as a library, as it's output changes unpredictably depending on which exact version of other dependencies are installed.

This is not a duplicate of https://github.com/koxudaxi/datamodel-code-generator/issues/440 .

**To Reproduce**

Example schema:
```json
    {
        "title": "SomeModel",
        "type": "object",
        "properties": {
            "attribute": {
                "title": "Attribute",
                "enum": [
                    1
                ],
                "type": "integer"
            }
        },
        "required": [
            "attribute"
        ]
    }
```

Used commandline:
```
$ datamodel-codegen --input file.json --enum-field-as-literal=one
```

**Expected behavior**
I expected the result to look something like
```
class SomeModel(BaseModel):
    attribute: Literal[1] = Field(..., title='Attribute')
```
instead it looks like
```
class SomeModel(BaseModel):
    attribute: Literal['1'] = Field(..., title='Attribute')
```

**Version:**
 - OS: Linux
 - Python version: 3.8.0
 - datamodel-code-generator version: 0.11.16

**Additional context**
The problem seems to lie in https://github.com/koxudaxi/datamodel-code-generator/blob/e2dcb199fc6da3c22aa5df4dd209721f1e71507e/datamodel_code_generator/types.py#L78 

Python caches specified generics - see also https://bugs.python.org/issue45679 -, which means that if
```
List[Union[str, int]]
```
was used in some dependency _before_ python parses this part, `List[Union[int, str]]` magically becomes `List[Union[str, int]]`. This is turn makes pydantic parse `[1]` to `['1']`. Whether or not `List[Union[str, int]]` was parsed by python before parsing `types.py` depends on the exact version of the dependencies which are installed.

For an example of this type caching, the following code runs without error in python 3.8:
```
from typing import List, Union

List[Union[str, int]]

assert str(List[Union[int, str]]) == "typing.List[typing.Union[str, int]]"
```
For how this can confuse pydantic, also the following code runs without error in python 3.8 with pydantic version 1.9.0:
```

from typing import List, Literal, Union

List[Union[str, int]]

class SomeModel(BaseModel):
    literals: List[Union[int, str]]

my_instance = SomeModel(literals=[1])

assert type(my_instance.literals[0]) == str
```
See also the warning in https://pydantic-docs.helpmanual.io/usage/types/#unions

duesenfranz:
Fixes #688

Before, dependening on the exact version of libraries imported before
parsing `types.py`, literal integers would get converted to strings,
as python caches specified generics and considers
`Union[int, str] == Union[str, int]` to be true.

### DIFF
--- a/datamodel_code_generator/types.py
+++ b/datamodel_code_generator/types.py
@@ -75,7 +75,7 @@ class DataType(_BaseModel):
     is_dict: bool = False
     is_list: bool = False
     is_custom_type: bool = False
-    literals: List[Union[int, str]] = []
+    literals: 'List[Union[int, str]]' = []
     use_standard_collections: bool = False
     use_generic_container: bool = False
     alias: Optional[str] = None
--- a/tests/data/expected/main/main_openapi_enum_models_all/output.py
+++ b/tests/data/expected/main/main_openapi_enum_models_all/output.py
@@ -16,6 +16,7 @@ class Pet(BaseModel):
     tag: Optional[str] = None
     kind: Optional[Literal['dog', 'cat']] = None
     type: Optional[Literal['animal']] = None
+    number: Literal[1]
 
 
 class Pets(BaseModel):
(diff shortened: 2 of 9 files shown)


duesenfranz:
@koxudaxi thanks for maintaining this! :) Could you release another version after merging this? :)

PR #704, opened 2022-02-04
fix booleans literals being converted to int

koonpeng:
### Bug fix

When using `--enum-field-as-literal`, a boolean enum is being converted to a `Literal[1]` instead of a `Literal[True]`. The cause of the bug is because the literals field does not specify bool as a valid type https://github.com/koxudaxi/datamodel-code-generator/blob/eb8210e9ea5bbb1a9dc56fb5b200b3a6a4387903/datamodel_code_generator/types.py#L78

The fix is to simply add `StrictBool` to the union, `StrictBool` is used because pydantic automatically converts data type whenever possible, putting `List[Union[bool, int, str]]` would cause integer enums of 0 or 1 to be converted to boolean, and `List[Union[int, bool, str]]` would cause the opposite.

### DIFF
--- a/datamodel_code_generator/types.py
+++ b/datamodel_code_generator/types.py
@@ -19,7 +19,7 @@ from typing import (
     Union,
 )
 
-from pydantic import create_model
+from pydantic import create_model, StrictBool
 
 from datamodel_code_generator import Protocol, runtime_checkable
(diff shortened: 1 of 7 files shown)


koxudaxi:
@koonpeng 
I have released the new version `0.11.19` 🎉

PR #726, opened 2022-03-09
no use constr when field-constraints

### LINKED ISSUE #724: Optional command to generate code without constr

The inclusion of `constr` in the generated datamodel when it contains regular expressions can cause the generated code to fail checks in mypy and flake8.

Specifically, with `https://github.com/compose-spec/compose-spec/blob/master/schema/compose-spec.json` as the target file, the following command attempts to generate the code.
````
datamodel-codegen --input compose-spec.json --input-file-type jsonschema --output model.py --field-constraints --target-python-version 3.8
````

Part of the generated result contains the following code
```
    services: Optional[Dict[constr(regex=r'^[a-zA-Z0-9._-]+$'), Service]] = None
````
This fails on mypy and flake8 checks.

The reason for the failure seems to be the use of constr, so I would like an optional command that outputs in a format that does not contain constr (nor conint).

If there is already an optional command that accomplishes this kind of functionality, it would be great to know.
I have already tried the `--field-constraints`, `-use-annotated` option commands, etc., but they did not solve this problem.

**Additional context**

If it is difficult to express a constraint by regular expression without using constr, I would like to see a behavior that replaces constr with str and outputs a warning.

Thank you for creating such a great package.

linshoK:
close #724 

I tried to prevent constr from appearing as a key type in Dict when `field_constraints == True`.


koxudaxi:
@linshoK 
I have added the unittest.
The PR removes validation logic for the dict's key pattern.
We should add the logic as a validator function in the future.

PR #765, opened 2022-05-11
Add Python 3.10 to PythonVersion

sgaist:
This pull request adds Python 3.10 to the PythonVersion enum so that it can also be constructed for it.

There's also a simple test that has been added to ensure the enum covers the version of Python being tested.

The test has been put in its own file following the model of `test_imports.py`.

The test is not parametrized as it goals is to fail if a new version of Python is tested but has not yet been added to the enum.


sgaist:
@koxudaxi you're welcome and thank you for the new release !

PR #789, opened 2022-06-08
Added IPv4/IPv6Network support.

ngaranko:
Added support for `ipv4-network` and `ipv6-network` formats to `string` type.


koxudaxi:
@ngaranko
I'm sorry for my late reply
LGTM
Thank you very much 😄

Issue #805, opened 2022-07-11
Constraint float casted into integer


FredericOutsight:
**Describe the bug**
When using alias maximum / minimum ... the generated type is Union[int, float, None]. As the class ConstaintsBase does not uses the BaseModel option smart_union, if a float is used, it will be cased into int by pydantic and the generated extremum will the erroneous.  

**To Reproduce**

If you patch the api_constrained.yaml that way
```diff --git a/tests/data/openapi/api_constrained.yaml b/tests/data/openapi/api_constrained.yaml
index 6340a51..4794135 100644
--- a/tests/data/openapi/api_constrained.yaml
+++ b/tests/data/openapi/api_constrained.yaml
@@ -169,6 +169,12 @@ components:
               - integer
             minimum: 1
             maximum: 1000
+          iq:
+            type: number
+            format: float
+            minimum: 0.2
+            maximum: 149.5
+
 
     Id:
       type: string
```
You can see that the generated code will be
`iq: Optional[float] = Field(None, ge=0.0, le=149.0)`
while:
`iq: Optional[float] = Field(None, ge=0.2, le=149.5)`
is expected

**Expected behavior**
Keep the actual float values for the keys : ge, le, gt, lt instead of converting them to integer and then back to float that create erroneous values

**Version:**
 - OS: Ubuntu 20.04
 - Python version: 3.8
 - datamodel-code-generator version: 1.12.0

**Additional context**
NA

koxudaxi:
The problem is fixed in the latest version. by the PR https://github.com/koxudaxi/datamodel-code-generator/issues/993

PR #806, opened 2022-07-11
Fix/float constraints cast as int

FredericOutsight:
#805 
Use BaseModel smart_union option for ConstrainsBase in order to avoid erroneous casts
Add related unit tests


koxudaxi:
I'm sorry for my late reply.
we should provide the option to add the smart_union option to keep current behavior.

koxudaxi:
Fied by #993
Thank you

PR #818, opened 2022-08-08
added option --use-double-quotes

nesb1:
Hello! Datamodel-codegen is a great utility. But i need so much double quotes in generated code because I use datamodel-codegen without --output option (due to the specifics of my task, i create python module myself) and i can`t specify skip-string-normalization using black config. And so i have quotes conflict.

For backward compability double-quotes by default is False.


docs/formatting.md
```diff
@@ -3,7 +3,10 @@
 Code generated by `datamodel-codegen` will be passed through `isort` and
 `black` to produce consistent, well-formatted results. Settings for these tools
 can be specified in `pyproject.toml` (located in the output directory, or in
-some parent of the output directory).
+some parent of the output directory). Also for black you can disable 
+skip-string-normalization with using datamodel-codegen option `--double-quotes`, 
+it will override your black config skip-string-normalization value.
+Using --double-quotes may be useful if you can`t use black config.
```
pawelrubin:
```suggestion
Using --double-quotes may be useful if you can't use black config. 
```

datamodel_code_generator/parser/openapi.py
```diff
@@ -184,6 +184,7 @@ def __init__(
         use_annotated: bool = False,
         use_non_positive_negative_number_constrained_types: bool = False,
         original_field_name_delimiter: Optional[str] = None,
+        double_quotes: bool = False,
```
pawelrubin: Here and in other places.

```suggestion
        use_double_quotes: bool = False,
```

nesb1:
@koxudaxi review please:)

koxudaxi:
@nesb1 @pawelrubin 
I have released the PR as 0.13.1
Thank you very much 🙇

luminoso:
Does this work when providing options via `pyproject.toml` ?

like so:

```
[tool.datamodel-codegen]
use-double-quotes = true
```

Can someone confirm that it isn't working? For me it is reacting to cli arguments only.

PR #840, opened 2022-09-02
Fix none related issues

AntoineD:
For a schema with a property that can be null and has a null default value, the pydantic model is not correctly generated.
For instance:
```json
{
  "properties": {
    "name": {
      "type": ["null", "string"],
      "default": "null"
    },
  },
  "type": "object"
}
```
is converted to
```python
class Model(BaseModel):
    name: Optional[Optional[str]] = 'null'
```
instead of
```python
class Model(BaseModel):
    name: Optional[str] = None
```

This PR addresses the default value conversion and the duplicated `Optional`.

### DIFF
--- a/datamodel_code_generator/model/base.py
+++ b/datamodel_code_generator/model/base.py
@@ -68,19 +68,19 @@ class DataModelFieldBase(_BaseModel):
 
         if not type_hint:
             return OPTIONAL
-        elif self.nullable is not None:
-            if self.nullable:
+        if self.nullable is not None:
+            if self.nullable and not type_hint.startswith('Optional'):
                 if self.data_type.use_union_operator:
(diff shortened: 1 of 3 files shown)


AntoineD:
Hi @koxudaxi,
it looks like you've rebased this PR, the failing CI jobs are due to codecov.
Do you need me to do something?

AntoineD:
@koxudaxi 

> Btw, How do we set `"null"` string value as default?

I do not know.

> Also, Can we use `"default": null` for default?
I would say yes in a json file.

koxudaxi:
@AntoineD 
>> Btw, How do we set "null" string value as default?
> I do not know.
We should not break current behavior.
I think this is not a bug.

koxudaxi:
no update

Issue #714, opened 2022-02-15
Error parsing external relative ref that contains a local ref


kepler:
**Describe the bug**
Parsing fails with a `KeyError` when processing an external ref that contains a local ref.

```
/datamodel_code_generator/parser/jsonschema.py", line 54, in get_model_by_path
    return get_model_by_path(schema[keys[0]], keys[1:])
KeyError: 'model_a'
```

**To Reproduce**

With two schemas in separate paths, which are to be processed separately but can reference each other:
File `model_a/types.openapi.yaml`:
```yaml

 ---
openapi: 3.0.3
info:
  title: Model A definitions
  version: "1.0"

paths: {}


components:
  schemas:

    model_a.input:
      type: object
      properties:
        name:
          type: string

    model_a.output:
      type: object
      properties:
        output:
          type: string
        input:
          $ref: "#/components/schemas/model_a.input"

```

Running the following works as expected:
```
$ datamodel-codegen --input model_a --output tmp/
```

Now consider a second file, `model_b/module.openapi.yaml`, that references `model_a.input`:
```yaml

---
openapi: 3.0.3
info:
  title: Model B definitions
  version: "1.0"

paths: {}


components:
  schemas:
    modules.quality_evaluation.QualityEvaluationRequest:
      type: object
      properties:
        input:
          $ref: "../model_a/types.openapi.yaml#/components/schemas/model_a.input"

```

Running the following also works as expected:
```
$ datamodel-codegen --input model_b --output tmp/
```

But now, if instead we ref to `model_a.output`, which contains a ref itself:
```yaml

---
openapi: 3.0.3
info:
  title: Model B definitions
  version: "1.0"

paths: {}


components:
  schemas:
    modules.quality_evaluation.QualityEvaluationRequest:
      type: object
      properties:
        input:
          $ref: "../model_a/types.openapi.yaml#/components/schemas/model_a.output"

```

Then running:
```
$ datamodel-codegen --input model_b --output tmp/
```
breaks the following:
```
Traceback (most recent call last):
  File "/VENV/lib/python3.9/site-packages/datamodel_code_generator/__main__.py", line 495, in main
    generate(
  File "/VENV/lib/python3.9/site-packages/datamodel_code_generator/__init__.py", line 353, in generate
    results = parser.parse()
  File "/VENV/lib/python3.9/site-packages/datamodel_code_generator/parser/base.py", line 426, in parse
    self.parse_raw()
[...]

PR #875, opened 2022-10-20
fix: parsing external relative ref that contains a local ref

Niraj-Kamdar:
Closes: #714


koxudaxi:
@Niraj-Kamdar 
I would review your PR.
Could you please revert the package name?

koxudaxi (changes requested):
Could you please revert the package name?

Niraj-Kamdar:
Ah sorry for that, I kept it same as where I deployed the fixed package but forgot to revert it back while opening PR, will make the changes.

koxudaxi:
@Niraj-Kamdar 
Thank you for pushing it.
But, Your commitment broke the existing unit test. 
Could you fix it?

koxudaxi:
@Niraj-Kamdar 
In the current master branch, the problem is fixed.
https://github.com/koxudaxi/datamodel-code-generator/pull/964

Thank you very much

PR #887, opened 2022-10-27
Drop python3.6 support

koxudaxi:


gjcarneiro:
I am working on a code base that runs  onpython 3.6 still, and will not upgrade for some time.  Is the intention to have the code generator itself require python >= 3.7 but generates code that still works on python 3.6?  Or is the generated code no longer going to support 3.6?

koxudaxi:
@gjcarneiro 
> Is the intention to have the code generator itself require python >= 3.7
Some libraries drop Python 3.6 support.
I don't want to keep complex dependency.

Also, I checked [the download list](https://pypistats.org/packages/datamodel-code-generator).
I think Python 3.6 users are decreasing now.

> but generates code that still works on python 3.6? Or is the generated code no longer going to support 3.6?

I would remove string annotation features like `Optional['OriginalModel']` for Python3.6

But, If a few users need Python 3.6 support.
I will close the PR.

gjcarneiro:
Would be great if you could generate code still compatible with Python 3.6, even if it needs to run on Python >= 3.7 now.  But it's your call.  Thanks!

Issue #892, opened 2022-11-02
Parser generates invalid import statement


antxln:
**Describe the bug**
Parser generates invalid code ```from ..Common. import Models```
Throws ```black.parsing.InvalidInput: Cannot parse: 7:15: from ..Common. import Models```

**To Reproduce**

Example schema:
```json

{
  "components": {
    "schemas": {
      "Public.Api.Models.ViewModel": {
        "type": "object",
        "properties": {
          "status": {
            "$ref": "#/components/schemas/Public.Common.Models.Status"
          },
        },
      },
      "Public.Common.Models.Status": {
        "enum": [
          "status"
        ],
        "type": "string"
      },
      "Public.Api.Models.V2.ViewModel": {
        "type": "object",
        "properties": {
          "id": {
            "type": "integer",
            "format": "int32"
          },
        },
      },
    },
  },
}

```

Used commandline:
```
$ datamodel-codegen --input openapi.json --input-file-type openapi --output model
```

**Expected behavior**

**Version:**
 - OS: Windows 10
 - Python version: 3.8.1
 - datamodel-code-generator version: 0.13.4

**Additional context**

PR #893, opened 2022-11-02
Fix generating import statement code

antxln:
Reference: #892

Code generator is appending a dot at end of module name in import statement, which is throwing an error with black. Maybe the dot can be prepended since it is indicating path heirarchy levels?


koxudaxi:
@antxln 
I have released a new version as `0.13.5`

PR #923, opened 2022-11-15
Look for local files with .json/yaml suffixes

crembo-dg:
When referring to local schemas, this change first checks whether the schema file exists. If it does not exist, it tries to find a similar file that has either a `.json` or `.yaml` suffix. If a file is found it is used instead of the missing suffix-less schema file.


koxudaxi:
@crembo-dg 
> this change first checks whether the schema file exists. If it does not exist, 

Is it a valid schema? :thinking: 
Does the behavior fill the JSON Schema's Specification?

crembo-dg:
> @crembo-dg Thank you for creating the PR.
> 
> > this change first checks whether the schema file exists. If it does not exist,
> 
> Is it a valid schema? 🤔 Does the behavior fill the JSON Schema's Specification?

You are right, this is not according to specs.

PR #924, opened 2022-11-16
Add support for tags and include responses without content

Aedial:
The exact implementation of responses without content is still to be decided


Aedial:
There are 2 versions of `parse_responses()`, the legacy one that doesn't return any empty response and `parse_responses_with_none()`, which returns the empty responses alongside the others.
I ran the tests, and both seem to work against the test suite, but some dependents might rely on the specific behavior of the legacy one.
Adding `include_empty_models directly` to `parse_responses()` would have been tempting, but it would break the signature that seems to define the `parse_*()` functions (and breaks inheritance on dependents).

Aedial:
Merging this PR would solve https://github.com/koxudaxi/fastapi-code-generator/issues/114, https://github.com/koxudaxi/fastapi-code-generator/issues/176 and https://github.com/koxudaxi/fastapi-code-generator/issues/207 (if I understood these correctly).
I already have working code for these issue, waiting for this PR to go through.

koxudaxi:
@Aedial
Should we add the option to call the new methods?

Aedial:
@koxudaxi  So, if I understand well, the "right" way would be to add another parameter to OpenAPIParser (datamodel_code_generator/parser/openapi.py#L244), generate (datamodel_code_generator/__init__.py#L200), and arg_parser (datamodel_code_generator/__main__.py#L349) ? I was fearing contributing to uncontrolled organic growth going this route, but it seems it is fine by you.

Aedial:
> Do we need tags' logic in datamodel-code-generator? We can write the method in fastapi-code-generator.

Putting the tags handling in fastapi-codegen makes things awkward, as it would mean putting a special clause for tags in parse_operation (fastapi-code-generator/parser.py), while keeping it in datamodel-code-generator stays consistent with the rest (parse_parameters, parse_request_body, parse_responses) and keeps the logic of Operation in one place. From a "separation of responsibilities" point of view, it feels cleaner.

Aedial:
Just realized I never committed the tests (oops)

koxudaxi:
@Aedial 
I'm sorry for my late reply.
I misunderstood. We don't need the new arguments of CLI.
We should add the argument for the parser.

PR #927, opened 2022-11-20
Fix deprecation warnings around "copy_on_model_validation"

Dominic-Walther:
Reduces Pytest warnings from 6145 down to 4.

### DIFF
--- a/datamodel_code_generator/reference.py
+++ b/datamodel_code_generator/reference.py
@@ -92,7 +92,7 @@ class Reference(_BaseModel):
     class Config:
         arbitrary_types_allowed = True
         keep_untouched = (cached_property,)
-        copy_on_model_validation = False
+        copy_on_model_validation = 'none'
 
     @property
     def short_name(self) -> str:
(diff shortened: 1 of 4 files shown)


koxudaxi:
@Dominic-Walther 
I guess the braking changes came from the PR.
https://github.com/pydantic/pydantic/pull/4093

We should change the value by the pydantic version.

Dominic-Walther:
Good catch.
Would be great to have an automatic check to verify compatibility with older versions of dependencies, as my commit would have likely broken something, despite passing all tests. Checking all possible version combinations would lead to exponential (if not factorial) growth in the number of required tests, so it's probably best to just check the oldest supported version of all direct dependencies. I'm not aware of a simple way of doing this however - maybe a poetry plugin? (Related issues: https://github.com/pypa/flit/issues/396, https://github.com/pdm-project/pdm/issues/906)

koxudaxi:
I agree with you.
We should check the oldest dependence.
I want to avoid a lot of unit test on CI :sweat:

Issue #820, opened 2022-08-10
Referenced basic types (like strings) are converted to classes


Trolldemorted:
**Describe the bug**
If a string is referenced, it is turned into a separate class that wraps the string.

**To Reproduce**

Given this schema:
```json
{
  "openapi" : "3.0.0",
  "components" : {
    "schemas" : {
      "UserId" : {
        "type" : "string",
      },
      "Tweet" : {
        "type" : "object",
        "properties" : {
          "author_id" : {
            "$ref" : "#/components/schemas/UserId"
          }
        }
      }
    }
  }
}
```
datamodel-code-generator emits:
```python
# generated by datamodel-codegen:
#   filename:  minimal.json
#   timestamp: 2022-08-10T13:22:26+00:00

from __future__ import annotations

from typing import Optional

from pydantic import BaseModel, constr


class UserId(BaseModel):
    __root__: str


class Tweet(BaseModel):
    author_id: Optional[UserId] = None
``` 
Would it be difficult to directly use a string in those cases?

**Version:**
 - OS: Ubuntu 20.04
 - Python version: 3.8.10
 - datamodel-code-generator version: 0.13.0

koxudaxi:
@Trolldemorted 
The CLI doesn't provide a solution.
We should implement a new option.
What do you expect?

```
class Tweet(BaseModel):
    author_id: Optional[str] = None
 ```
Or,
```
UserId = str
class Tweet(BaseModel):
    author_id: Optional[UserId] = None
 ```
🤔

Trolldemorted:
The first one looks way better, keeping things simple is always good :)

Dominic-Walther:
This works fine for single references, but if a definition is referenced multiple times it causes all field attributes (e.g. `Annotated[str, Field(regex='...', description='...')]`) to be duplicated unnecessarily for each field. This makes keeping the output model internally consistent more difficult when editing it manually.

While i think this change has some uses, I'd prefer a choice between "force inline", "force class" and potentially "auto", which would check the reference count to decide on a case by case basis.

anneum:
@koxudaxi Thank you for implementing the feature. With the given test scheme everything works as expected.
We tried the slightly larger [Twitter API specification](https://api.twitter.com/2/openapi.json) as source and get the following error:
```python
Traceback (most recent call last):
  File "/home/username/.local/lib/python3.8/site-packages/datamodel_code_generator/__main__.py", line 626, in main
    generate(
[...]

PR #933, opened 2022-12-06
Add collapse root model feature

### LINKED ISSUE #638: Referenced string is not real string (cannot be used as dict key, doesn't equal to string)
**Describe the bug**

**To Reproduce**

Example schema:
```json
openapi: "3.0.0"
info:
  version: 1.0.0
  title: File endpoint
paths:
 get_file:
    parameters:
      - $ref: "#/components/schemas/FileRequest"

components:
  schemas:
    FileHash:
      type: string
      min_length: 32
      max_length: 32
      regex: "^[a-fA-F\d]{32}$"
    FileRequest:
      type: object
      required:
        - file_hash
      properties:
        file_hash:
          $ref: "#/components/schemas/FileHash"
```

Used commandline:
```sh
$ datamodel-codegen file.yml
```

Generated python code:
```py
class FileHash(BaseModel):
    __root__: str = Field(..., max_length=32, min_length=32, regex="^[0-9a-fA-F]+$")

class FileRequest(BaseModel):
    file_hash: FileHash
```

Python usage
```py
file_hash = "a"*32
f = FileRequest(file_hash=file_hash)

assert f.file_hash == file_hash) # does not equal

d = dict()
d[f.file_hash] = "asdf" # throws TypeError: unhashable type: 'FileHash'
```

**Expected behavior**
- original string equals with field
- can be used as key in dict

**Version:**
 - OS: linux
 - Python version: 3.9
 - datamodel-code-generator version: 0.11.14

i404788:
Partially resolves #638 #820, as written it may not cover all cases/features (annotations/fields might not be carried over afaik).
Also makes use of PEP 572 for clarity which might not be desirable if older version of python need to be supported.

If there is interest in this upstream I'll try to resolve the issues described above.


datamodel_code_generator/parser/base.py
```diff
@@ -692,6 +694,14 @@ def parse(
                 for duplicate in duplicates:
                     models.remove(duplicate)
 
+            if self.collapse_root_models:
+                for model in models:
+                    for model_field in model.fields:
+                        if ref := model_field.data_type.reference:
```
koxudaxi: We support Python 3.7 yet :)

koxudaxi:
@i404788
I have released the new version as 0.14.1 :tada:

PR #938, opened 2022-12-09
Fix indents for multi-line docstrings

Dominic-Walther:
Docstrings spanning multiple lines now get properly indented.
I added a copy of api.yaml with multi-line descriptions to avoid making changes to 28 other tests that depend on api.yaml.


koxudaxi:
I'm sorry for my late reply.
Looks good :)

PR #962, opened 2022-12-23
Avoid field name beginning with an underscore

### LINKED ISSUE #917: @ symbol in JSON keys produces private variable
**Describe the bug**
We have JSON that has `@id` and `@type` which are reflected in the schema below. When we generate the python model using the datamodel-codegen library it produces variables that we can't access. This is likely due to [pydantic hiding variables](https://pydantic-docs.helpmanual.io/usage/models/#automatically-excluded-attributes) starting with an underscore. 

**To Reproduce**

Example schema:
```json
{
    "$id": "schema_v2.json",
    "$schema": "http://json-schema.org/schema#",
    
    "type": "object",
    "properties": {
        "@id": { 
            "type": "string",
            "format": "uri",
            "pattern": "^http.*$",
            "title": "Id must be presesnt and must be a URI"
        },    
        "@type": { "type": "string" },
        "profile": { "type": "string" }
    },
    "required": ["@id", "@type"]
} 
```

Produces the following model:
```
class Model(BaseModel):
    _id: AnyUrl = Field(..., alias='@id', title='Id must be presesnt and must be a URI')
    _type: str = Field(..., alias='@type')
    profile: Optional[str] = None
```

and if we try and use it we can't access id or type:
```
$ python
Python 3.9.10 (main, Jan 15 2022, 11:48:04) 

>>> from model2 import Model
>>> v2 = Model(_id="http://example.com/iiif/2", _type="service")
>>> print (v2)
profile=None
```

Used commandline:
```
$ datamodel-codegen  --input schema_v2.json --input-file-type jsonschema --output model2.py
```

**Expected behavior**
Ideally we would look the following conversion where the _ is removed:

```
class Model(BaseModel):
    id: AnyUrl = Field(..., alias='@id', title='Id must be presesnt and must be a URI')
    type: str = Field(..., alias='@type')
    profile: Optional[str] = None
```

**Version:**
 - OS:  Mac OS X Monterey 12.6
 - Python version:  3.9.10
 - datamodel-code-generator version: 0.13.5

**Additional context**
JSON keys `@id`, `@type` and `@context` are used in [JSON-LD](https://json-ld.org/)

ronlib:
fixes #917

Adds a `field` prefix in any situation where the generated field name begins with `_`, which makes pydantic consider it by default as a private field.


ronlib:
I can see it broke some tests. Let me know what you think about the new behavior I introduced here

koxudaxi:
@ronlib
I don't like to break current behavior.
But, JSON Schema is a public interface and should not be private.

We should add a CLI option to change `field_` prefix to another string  after merging the PR.

khvn26:
This breaks other output types such as `typing.TypedDict`.

PR #968, opened 2022-12-27
Fix allOf with common prefix and $ref

azatoth:
Per #714 test is needed to examplify the issue.

### DIFF
--- a/datamodel_code_generator/parser/jsonschema.py
+++ b/datamodel_code_generator/parser/jsonschema.py
@@ -516,6 +516,7 @@ class JsonSchemaParser(Parser):
             )
         # ignore an undetected object
         if ignore_duplicate_model and not fields and len(base_classes) == 1:
+            self.model_resolver.delete(path)
             return self.data_type(reference=base_classes[0])
         if self.use_title_as_name and obj.title:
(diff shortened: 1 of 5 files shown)


koxudaxi:
@azatoth 
I added some lines to fix the problem.

azatoth:
> @azatoth Thank you for creating this PR. I added some lines to fix the problem.

Thank you :) I didn't know what the underlying issue was, but I thought at least I could add a test to exemplify the issue.

koxudaxi:
@azatoth 
I explain the problem.
1. the code generator wants to create a new class inherited allOf's schemas.
2. The class name is made with a field name. `foo_bar` ->  `FooBar`
3.  the case doesn't create a new class because the schema is one.
4. The generator decides the class name of `FooBar` schema 
5. `FooBar` is reserved as the same name as another schema name. 
6. The generator changes the name to `FooBar1`

azatoth:
@koxudaxi Thanks for the explanation; now it makes perfectly clear what the underlying issue was.

koxudaxi:
@azatoth 
I have released the new version as 0.14.1 :smile:

Issue #763, opened 2022-05-09
datamodel-codegen creates models with duplicate fields


pritishperpetua:
**Describe the bug**

- I am trying to auto-generate schema models using datamodel-codegen on an openapi documentation:- https://d3a0d0y2hgofx6.cloudfront.net/openapi/en-us/dsp/3-1/openapi.yaml
- The generated schema models has duplicate fields with incorrect typehints. 

**To Reproduce**

Used command line:
```
$ curl https://d3a0d0y2hgofx6.cloudfront.net/openapi/en-us/dsp/3-1/openapi.yaml -o amz_openapi.yaml
$ swagger-cli bundle --dereference amz_openapi.yaml > amz_openapi_dereferenced.yaml
$ datamodel-codegen --reuse-model --target-python-version 3.9 --use-default --input amz_openapi_dereferenced.yaml  --output amz_api_schema.py
```

**Expected behavior**
- Model `OrderV22` has `currencyCode` as a string field. 

**Observed behavior**
- Model `OrderV22` has multiple fields for `currencyCode`

<img width="1768" alt="image" src="https://user-images.githubusercontent.com/61429653/167482207-834022c8-62b3-46d6-8f56-8193bd89e667.png">

Similar behavior can also be observed for `LineItemV31`

Any insights on what might me going wrong here would be really appreciated. 

**Version:**
 - OS: macOS
 - Python version: 3.9
 - datamodel-code-generator version: 0.12.0

koxudaxi:
@pritishperpetua  
I have released the PR as `0.15.0`
Thank you very much!!

dannyrohde:
I believe this change introduces a problem for generating a Pedantic model from an Amazon JSON source.

In the past, the library generated a list of properties on the class, with the *last* one being correct:

```
    fulfillment_availability: Optional[List[Any]] = None
    fulfillment_availability: Optional[List[Any]] = None
    fulfillment_availability: Optional[List[Any]] = None
    fulfillment_availability: Optional[List[Any]] = None
    fulfillment_availability: Optional[List[Any]] = None
    fulfillment_availability: Optional[List[Any]] = None
[…]
    fulfillment_availability: Optional[List[FulfillmentAvailabilityItem]] = Field([…])
```

[...]

PR #974, opened 2022-12-30
Ignore duplicate name field on allOf

koxudaxi:
Closes: https://github.com/koxudaxi/datamodel-code-generator/issues/763


dannyrohde:
Not sure how to open a regression bug on this, so will try to comment here. Let me know if that's not the right way to deal with this.

This change introduces a regression bug for me. I have a similar situation as the original bug, with the difference that the *last* declaration is correct instead of the first.

I've explained it more in a comment on the original bug: https://github.com/koxudaxi/datamodel-code-generator/issues/763#issuecomment-1448420819

Issue #993, opened 2023-01-06
Incorrect constraint definition with --field-constraints option


ninowalker:
**Describe the bug**

When ` --field-constraints` is enabled, the generated code produces incorrect bounds on a number field. Specifically it rounds the bounds to integers!

This error only occurs with this option enabled.
 
**To Reproduce**

Example schema:
```json
{
  "additionalProperties": false,
  "properties": {
    "bounded_number": {
      "description": "Bounded between -1.2 and 1.2",
      "maximum": 1.2,
      "minimum": -1.2,
      "type": "number"
    }
  },
  "required": [],
  "type": "object"
}
```

Output:
```python
class Model(BaseModel):
    ...

    bounded_number: Optional[float] = Field(None, ge=-1.0, le=1.0) ######### < SEE BUG HERE!
    """
    Bounded between -1.2 and 1.2
    """
```

Used commandline:
```
$ datamodel-codegen --version
0.15.0
$ datamodel-codegen --input bug.json \
        --input-file-type jsonschema \
        --base-class pydantic.BaseModel \
        --allow-population-by-field-name \
        --disable-timestamp \
        --use-schema-description \
        --use-field-description \
        --use-title-as-name \
        --field-constraints \
        --collapse-root-models \
        --reuse-model \
        --use-default --output /dev/stdout
```

**Expected behavior**
The bounds should not be rounded. 

**Version:**
 - OS: Linux
 - Python version: 3.10.1
 - datamodel-code-generator version: 0.15.0

**Additional context**

Related to: https://github.com/koxudaxi/datamodel-code-generator/issues/583#issuecomment-1307357370

Compare the output without `--field-constraints`, which curries the correct values:

```python
class Model(BaseModel):
    ...
    bounded_number: Optional[confloat(ge=-1.2, le=1.2)] = None
    """
    Bounded between -1.2 and 1.2
    """
```

ninowalker:
Note @koxudaxi - this patch fixes my issue: https://github.com/ninowalker/datamodel-code-generator/commit/a8a59eaaf6cbb5f2fae2a35e12b19e058571e9a4

This ensures that the input to `_get_strict_field_constraint_value` is a float type which is then coerced to an int.

Here's updated inputs/outputs:

Input:
```json
{
  "additionalProperties": false,
  "properties": {
    "bounded_number": {
      "description": "Bounded between -1.2 and 1.2",
      "maximum": 1.2,
      "minimum": -1.2,
[...]

PR #1003, opened 2023-01-10
Fix field constraint number coersion bug

ninowalker:
Fixes https://github.com/koxudaxi/datamodel-code-generator/issues/993


koxudaxi:
@ninowalker 
Thank you very much!!

Issue #1027, opened 2023-01-18
Not `None` default values for arg_parse override settings in `pyproject.toml`


JensHeinrich:
**Describe the bug**
`collapse-root-models = true` is ignored, when set via `pyproject.toml` 

**To Reproduce**

Example schema:
[`error.schema`](https://raw.githubusercontent.com/folio-org/raml/69f6074f5f48fd58e379e9ae76c892aa68c115ea/schemas/error.schema) and the referenced [`parameters.schema`](https://raw.githubusercontent.com/folio-org/raml/69f6074f5f48fd58e379e9ae76c892aa68c115ea/schemas/parameters.schema)

Used commandline:
```
$ datamodel-codegen --url https://raw.githubusercontent.com/folio-org/raml/4aaeffa22d5461b41ac761870dcfce58bf8db387/schemas/error.schema  
```

**Expected behavior**
After adding the `collapse-root-models` entry to the `pyproject.toml` this should result in the same output as 
```
datamodel-codegen --collapse-root-models --url https://raw.githubusercontent.com/folio-org/raml/4aaeffa22d5461b41ac761870dcfce58bf8db387/schemas/error.schema  
```

**Version:**
 - OS: Linux
 - Python version: Python 3.10.7
 - datamodel-code-generator version: 0.16.0

**Additional context**

JensHeinrich:
What's your release policy for minor releases (basically: when can I update my project dependencies to include the fixed version^^)
 
 @koxudaxi

koxudaxi:
@JensHeinrich 
Thank you for creating the PRs 😄 

We don't have a policy.
But I avoid sending annoying dependency update notifications to users 
Recently, I released every about 1 ~ 2 weeks.
Don't you prefer to install it by using git?
`pip install git+https://github.com/koxudaxi/datamodel-code-generator.git`
Or, When do you want the fixed version?

JensHeinrich:
I am just including it as dev dependency via poetry, so a hard version would be useful, but I can install it by using git, just forgot this option^^

PR #1029, opened 2023-01-18
Fix/remove not none arg parse defaults

JensHeinrich:
This applies the same fixes as for #1027 to more options


koxudaxi:
@JensHeinrich 
I have released the new version as `0.16.1`
Thank you very much!!

PR #1034, opened 2023-01-19
Add --use-default-kwarg option for better type checker compatibility

zackyancey:
Python 3.11 introduced the `@dataclass_transform` decorator, and older
verisions of python can access the same from the `typing_extensions`
package. This decorator lets type checkers infer the right `__init__`
function for "dataclass-like" classes. It relies on some conventional
keyword arguments to determine whether a given field is a required
argument for `__init__`. Essentially, if the `default=` kwarg is
present, then the field is not required.

Pydantic allows for the `default` argument to be positional, but this
doesn't give the right hints to the type checker. If we set `default` as
a kwarg for fields that do have defaults, then the signature of
`__init__` is inferred correctly.

Example generated code:

```python

# Without --use-default-kwarg: type checkers will assume `b` and `c` are
# required arguments to `__init__`
class Model(BaseModel):
    a: str = Field(..., description='Required field')
    b: Optional[str] = Field(None, description='Optional field')
    c: str = Field("default", description='Another optional field')

# With --use-default-kwarg: type checkers will correctly assume `b` and
# `c` are optional arguments to `__init__`
class Model(BaseModel):
    a: str = Field(..., description='Required field')
    b: Optional[str] = Field(default=None, description='Optional field')
    c: str = Field(default="default", description='Another optional field')
```


koxudaxi:
@zackyancey  
I have released the new version as `0.16.1`
Thank you very much!!

PR #1053, opened 2023-01-26
Proposal for adding version metadata to header section

ghandic:


ghandic:
Could be an additional flag similar to the timestamp

Allows for

```
# generated by datamodel-codegen 0.16.1:
#   filename:  example.yml
#   timestamp: 2023-01-26T01:11:51+00:00
```

koxudaxi:
@ghandic
Looks good 😄 
But I don't want to affect existing users.
We should apply the changes with the new CLI option.

ghandic:
Agreed, just an initial PR to gauge appetite 😄

ghandic:
I've added it similar to the timestamp, lmk thoughts

ghandic:
Looks good ⭐  thanks!

koxudaxi:
@ghandic 
I have released a new version `0.17.0`
The version has the option.
Thank you very much!!

PR #1056, opened 2023-01-26
Add `$ref` support to schemas

ytoml:
Hi, I'm a newbie to OpenAPI and this generator.
This generator doesn't seem to have `$ref` resolving at `schemas`, thus I added.
I would help us keep Yamls separated and easier to manage.

### DIFF
--- a/datamodel_code_generator/parser/openapi.py
+++ b/datamodel_code_generator/parser/openapi.py
@@ -431,6 +431,8 @@ class OpenAPIParser(JsonSchemaParser):
                 path_parts = list(source.path.parts)
             with self.model_resolver.current_root_context(path_parts):
                 if OpenAPIScope.Schemas in self.open_api_scopes:
+                    if "$ref" in schemas:
+                        schemas = self._get_ref_body(schemas["$ref"])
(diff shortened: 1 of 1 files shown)


koxudaxi:
@ytoml 
We can't assign `Reference Object` to `schemas` object directly.
We should define `Reference Object` as the value of `Map`.

`schemas | Map[string, Schema Object | Reference Object]`
https://spec.openapis.org/oas/v3.0.0#fixed-fields-5

ytoml:
Thank you for information!
Why I thought it valid first was because `swagger-cli bundle` did expand `$ref` at `schemas`.  Then, `swagger-cli`'s implementation seems to be wrong...? (or it fits some legacy specification?).

ytoml:
Anyway, closing this PR, thanks!

koxudaxi:
I can't find the rule in the old specification.
https://github.com/OAI/OpenAPI-Specification/blob/main/versions/2.0.md

> swagger-cli's implementation seems to be wrong...?

`swagger-cli` may support json pointer anywhere :thinking:

PR #1083, opened 2023-02-03
OpenAPI - Add support for query parameter model generation

hambrosia:


pvicol:
@koxudaxi - would appreciate your guidance on this. I am a bit lost as to how the parses parses the data and creates the models. The current changes in this PR, parses data in a very similar way the `parse_responses` does, however, when inspecting in a debugger using PyCharm, I am noticing that the `full_name` is not getting set when using `self.parse_schema`. Any help into where the models are being created from parsed data would be greatly appreciated!

datamodel_code_generator/parser/openapi.py
```diff
@@ -274,7 +274,11 @@ def get_ref_model(self, ref: str) -> Dict[str, Any]:
 
     def parse_parameters(self, parameters: ParameterObject, path: List[str]) -> None:
         if parameters.name and parameters.schema_:  # pragma: no cover
-            self.parse_item(parameters.name, parameters.schema_, [*path, 'schema'])
+            data_model_type = self.parse_item(
+                parameters.name, parameters.schema_, [*path, 'schema']
+            )
+            if OpenAPIScope.Parameters in self.open_api_scopes:
+                self.results.append(data_model_type)
```
koxudaxi: @hambrosia  @pvicol 
`self.parse_item` returns `DataType`.But, `self.results` is the list of `DataModel`

koxudaxi:
@hambrosia @pvicol
Thank you for creating this PR and committing.
I made the parts of the path. [fastapi-code-generator]( https://github.com/koxudaxi/fastapi-code-generator) is need parts like path and parameter.
So, the parser only parses and doesn't create the model.
After parsing, fastapi-code-generator gets the parsed `DataType`.
But, you need models for your use-case.
We should add create field and model from parsed `DataType`.
I agree the code is not clear. We should re-factor the near future :sweat_smile: 
We can understand creating the model in `parser/jsonschema.py`

pvicol:
> @hambrosia @pvicol Thank you for creating this PR and committing. I made the parts of the path. [fastapi-code-generator](https://github.com/koxudaxi/fastapi-code-generator) is need parts like path and parameter. So, the parser only parses and doesn't create the model. After parsing, fastapi-code-generator gets the parsed `DataType`. But, you need models for your use-case. We should add create field and model from parsed `DataType`. I agree the code is not clear. We should re-factor the near future 😅 We can understand creating the model in `parser/jsonschema.py`

Thanks for doing this. After reviewing your changes, this makes so much more sense!
Definitely agree about refactoring, adding doc strings would also help!

koxudaxi:
@hambrosia @pvicol 
I have released a new version `0.17.1`
The version has an option `--openapi-scopes parameters`
Thank you very much!!

hambrosia:
Thank you @koxudaxi ! So fast!

PR #1187, opened 2023-03-19
Bugfix: with deeply nested modules: empty paths did not get an __init__.py file created

dataway:
Consider a model named `a.b.c.d.e` (as an example see tests/data/jsonschema/nested_skip.json)

The output created the directories `a/`, `a/b/` and `a/b/c/`, however the directories `a/` and `a/b/` should have an empty `__init__.py` file, which was not the case. This patch fixes this. A corresponding test case is also included.


koxudaxi:
@dataway 
I have released the new version, `0.17.2`.
Thank you very much!!

PR #1222, opened 2023-03-31
(🎁) deps: make typed-ast optional

KotlinIsland:
typed-ast is a very annoying package to deal with. This project doesn't even use it, so I assume it's only there to constrain old versions of black/mypy?

Context of transient constraints in poetry:
https://github.com/python-poetry/poetry/issues/4991


koxudaxi:
@KotlinIsland 
I have released the new version, `0.17.2`.
Thank you very much!!

KotlinIsland:
@koxudaxi This change didn't work and `typed-ast` is still mandatory 😥

Do you know why this is here to begin with? Seems to be something to do with black/mypy, both of which no longer depend on it. Perhaps we could just remove it entirely?

koxudaxi:
@KotlinIsland 
I build wheel and extract METADAT from the wheel file.
poetry defines `typed-ast` is required. :thinking: 
```
Requires-Dist: typed-ast (>=1.4.2); python_full_version < "3.9.8"
Requires-Dist: typed-ast (>=1.5.0); python_full_version >= "3.9.8"
```

PR #1246, opened 2023-04-15
(🎁) remove typed-ast dependency

KotlinIsland:
This is a follow up to #1222.

Typed-ast is a very painful dependency, I'm not sure why it is included in the dependencies to begin with. Modern versions of black and mypy no longer use typed-ast.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -1449,4 +1449,4 @@ http = ["httpx"]
 [metadata]
 lock-version = "2.0"
 python-versions = "^3.7"
-content-hash = "678e87f0717456439bed0c884b9d0a57baa08bf9de3e1b0e779e10992eeafff2"
+content-hash = "1cd7128d08b561c7f6d9b416b70fa3023c46250586c21d719af50aa59c7f6f47"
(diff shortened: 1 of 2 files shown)


koxudaxi:
@KotlinIsland 
I didn't write the detail of the PR/
I guess some old typed-ast doesn't support 3.9.8.
https://github.com/koxudaxi/datamodel-code-generator/pull/626
https://github.com/koxudaxi/datamodel-code-generator/pull/634

koxudaxi:
I think I added the dependencies for the problem.
https://stackoverflow.com/questions/69912264/python-3-9-8-fails-using-black-and-importing-typed-ast-ast3

KotlinIsland:
@koxudaxi Perhaps I don't understand every facet of this situation, but from what I can tell, certain versions of typed-ast are broken with certain versions of python or mypy/black. Unfortunately poetry doesn't support constraints of transitive dependencies. As far as I know, current tools do not use typed-ast. typed-ast doesn't have wheels for python 3.11, so therefor adds a complexity for anyone using it (need to install build tools to even install it). I think we should remove this dependency completely, if a downstream project is having issues with typed-ast, then they can add their own constraint for the version.

PR #1248, opened 2023-04-15
(🎁) log file type when `--input-file-type` is auto.

### LINKED ISSUE #1220: (🎁) Log input file type when when `--input-file-type` is `auto`
I was left a little confused when my json file was silently detected as jsonschema instead of json.

KotlinIsland:
This helps with the UX when trying to generate for a normal json file and it is automatically inferred as json schema.

- resolves #1220


KotlinIsland:
Maybe this could mention the `--input-file-type` option

PR #1259, opened 2023-04-18
(🎁) Include `py.typed` to mark project as typed

KotlinIsland:


koxudaxi:
@KotlinIsland 
I didn't know about PEP 561.
Thank you very much!!

Issue #1084, opened 2023-02-04
Response models not generated for paths that contains refs for nested schemas


pvicol:
**Describe the bug**
It has been observed that when attempting to generate responses for paths that reference nested schemas, the models are not returned.

**To Reproduce**

Example schema:
```yaml

openapi: "3.0.0"
info:
  version: 1.0.0
  title: Swagger Petstore
  license:
    name: MIT
  description: |
    This description is for testing
    multi-line
    description

servers:
  - url: http://petstore.swagger.io/v1
security:
  - BearerAuth: []
paths:
  /pets:
    get:
      summary: List all pets
      operationId: listPets
      parameters:
        - name: limit
          in: query
          description: How many items to return at one time (max 100)
          required: false
          schema:
            default: 0
            type: integer
            format: int32
        - name: HomeAddress
          in: query
          required: false
          schema:
            default: 'Unknown'
            type: string
        - name: kind
          in: query
          required: false
          schema:
            default: dog
            type: string
      responses:
        '200':
          description: A paged array of pets
          headers:
            x-next:
              description: A link to the next page of responses
              schema:
                type: string
          content:
            application/json:
              schema:
                type: array
                items:
                  $ref: "#/components/schemas/Pet"
        default:
          description: unexpected error
          content:
            application/json:
              schema:
                $ref: "#/components/schemas/Error"
components:
  schemas:
    Pet:
      type: object
      required:
        - id
        - name
      properties:
        id:
          type: integer
          format: int64
        name:
          type: string
        tag:
          type: string
    Error:
      required:
        - code
        - message
      properties:
        code:
          type: integer
          format: int32
        message:
          type: string

```

Above schema generates this code:
```python
from __future__ import annotations

from typing import Optional

from pydantic import BaseModel


class Pet(BaseModel):
    id: int
    name: str
    tag: Optional[str] = None


class Error(BaseModel):
    code: int
    message: str
```

Used commandline:
```
$ datamodel-codegen --input query_parameters.yaml --output model.py --openapi-scopes schemas paths
```

**Expected behavior**
The expected behavior is to create a model:

```python
from __future__ import annotations

from typing import Optional

from pydantic import BaseModel


class Pet(BaseModel):
    id: int
    name: str
    tag: Optional[str] = None


class Error(BaseModel):
    code: int
    message: str


class PetsGetResponse(BaseModel):
    id: int
    name: str
    tag: Optional[str] = None

```

The schema that does work to generated the expected models:
[...]

PR #1261, opened 2023-04-19
Fix OpenAPI array response

koxudaxi:
Related Issue: https://github.com/koxudaxi/datamodel-code-generator/issues/1084 https://github.com/koxudaxi/datamodel-code-generator/pull/1083


tests/data/expected/parser/openapi/openapi_parser_with_query_parameters/output.py
```diff
@@ -53,9 +53,7 @@ class PetsGetParametersQuery(BaseModel):
 
 
 class PetsGetResponse(BaseModel):
-    id: int
-    name: str
-    tag: Optional[str] = None
+    __root__: List[Pet]
```
koxudaxi, line 56: @pvicol @hambrosia
https://github.com/koxudaxi/datamodel-code-generator/pull/1083
I found the problem in the PR.
The current implementation doesn't create an array object on response.
But it creates an item of an array.

It means the PR breaks current behavior.

PR #1274, opened 2023-04-24
Import Annotated from typing_extensions for Python <= 3.8

airwoodix:
This patch fixes a backwards compatibility issue when targeting Python <= 3.8.

The [`typing.Annotated`](https://docs.python.org/3/library/typing.html#typing.Annotated) type was introduced only in Python 3.9, and should thus be imported from `typing_extensions` when `target-python-version` is `'3.8'` or less.


airwoodix:
@koxudaxi Thanks for fixing the unit test and the review! Sorry that I didn't find time to do it earlier.

koxudaxi:
@airwoodix 
I have released the PR as 0.19.0.

Issue #1289, opened 2023-04-30
Missed schema when jsonschema has allOf-anyOf combination.


itaru2622:
**Describe the bug**
Missed schema of property when jsonschema has allOf-anyOf combination.

**To Reproduce**

Example schema:

case1: all described in single definition (no $ref)
```json
{
  "definitions": {
    "Target": {
      "allOf": [
        {
          "anyOf": [
            {
              "type": "object",
              "properties": {
                "first": {
                  "type": "string"
                }
              },
              "required": [
                "first"
              ]
            },
            {
              "type": "object",
              "properties": {
                "second": {
                  "type": "string"
                }
              },
              "required": [
                "second"
              ]
            }
          ]
        },
        {
          "properties": {
            "third": {
              "type": "string"
            }
          },
          "required": [
            "third"
          ]
        }
      ]
    }
  }
}
```
case2:  definition with 'ref'
```json
{
  "definitions": {
    "First": {
      "properties": {
        "first": {
          "type": "string"
        }
      },
      "required": [
        "first"
      ]
    },
    "Second": {
      "properties": {
        "second": {
          "type": "string"
        }
      },
      "required": [
        "second"
      ]
    },
    "Target": {
      "allOf": [
        {
          "anyOf": [
            {
              "$ref": "#/definitions/First"
            },
            {
              "$ref": "#/definitions/Second"
            }
          ]
        },
        {
          "properties": {
            "third": {
              "type": "string"
            }
          },
          "required": [
            "third"
          ]
        }
      ]
    }
  }
}

```

Used commandline:
```
$ datamodel-codegen  --input-file-type jsonschema  --input file.json
```

**Generated result**

at case1:
```python
from __future__ import annotations
from pydantic import BaseModel

class Target(BaseModel):
    third: str
```

at case2:
```python
from __future__ import annotations
from pydantic import BaseModel

class First(BaseModel):
    first: str
class Second(BaseModel):
    second: str
class Target(BaseModel):
    third: str

```
**Expected behavior**
I expect, in both cases, Target model has inheritance of  First or Second model, something like below.

```python
:
class TargetAnyof(BaseModel):
    __root__: Union[First, Second]
class Target(TargetAnyof):
    third: str
```

When I tested just 'anyof' part (without allof), the result has Union[first,second] as  expeced. so, something wrong in processing the combination of allof and anyof, maybe in parser/jsonschema.py

**Version:**
 - OS: debian 11
[...]

PR #1291, opened 2023-05-01
Fix allOf-anyOf

koxudaxi:
Closes: https://github.com/koxudaxi/datamodel-code-generator/issues/1289


koxudaxi:
@itaru2622 
Can you check this PR?

itaru2622:
@koxudaxi  Thank you for your working on fixing.

I checked this PR branch (through test code), and found that: It is mostly fine, but ...

in tests/data/expected/main/main_all_of_any_of/direct.py and reference.py

1)  It misses the final model having inheritance of  Union[Target1 and Target2].
     Users can easily build the above final model themselves from the current result, but it should generate final model to provide results similar to other current behaviors.
2)   the above final model should have the name 'Target' which appeared in jsonschema. 
      for avoid users confused,  current 'Target' model having just  'third' attribute should be renamed such as 'TargetItem'.
      
I missed to show you the patch by myself to meet the above, since I have less experience of the code of this OSS yet.

in detail, about above comment1, when I checked anyOf parts, I got below result.
when the final model is not generated, the legacy behavior may be  broken.
```json
{
  "definitions": {
    "FirstInAnyof": {
      "properties": {
        "first": {
          "type": "string"
        }
      },
      "required": [
        "first"
      ]
    },
    "SecondInAnyof": {
      "properties": {
        "second": {
          "type": "string"
        }
      },
      "required": [
        "second"
      ]
    },
    "TargetForAnyofPart": {
      "anyOf": [
        {
          "$ref": "#/definitions/FirstInAnyof"
        },
        {
          "$ref": "#/definitions/SecondInAnyof"
        }
      ]
    }
  }
}
```
```python
:
class FirstInAnyof(BaseModel):
    first: str
class SecondInAnyof(BaseModel):
    second: str
class TargetForAnyofPart(BaseModel): # inherit one of First or Second
    __root__: Union[FirstInAnyof, SecondInAnyof]
```

itaru2622:
@koxudaxi  oops, this PR is going to be merge before applying the request of change described in my comment?

koxudaxi:
@itaru2622 
>  this PR is going to be merge before applying the request of change described in my comment?

No, We should clear all problems before merging.
>   inherit one of First or Second

What do you mean? Root model can't inheritance another base model.

koxudaxi:
> the above final model should have the name 'Target' which appeared in jsonschema.

make sense.

itaru2622:
@koxudaxi after some investigation, 

I found the preferable naminging logic maybe already implemented in the current code
somewhere within parse_any_of  => parse_list_item  => ... call stacks,
but not sure the exact code yet.

I can get following  classes (forcuse on just names of classes)
```python
:
class FirstInAnyOf(BaseModel):
    first: str
class SecondInAnyOf(BaseModel):
    second: str
class TargetItem(BaseModel): # the additional model has the suffix('Item' ) in name when it needs to be created.
    third: str
class Target(BaseModel): # The model keeps the same name as the jsonschema.
    __root__: Union[FirstInAnyOf, SecondInAnyOf, TargetItem]

```
by following jsonschema.
```
{
  "definitions": {
    "FirstInAnyOf": {
      "properties": {
        "first": {
          "type": "string"
        }
      },
      "required": [
        "first"
      ]
    },
    "SecondInAnyOf": {
      "properties": {
        "second": {
          "type": "string"
        }
      },
      "required": [
        "second"
      ]
    },
    "Target": {
      "anyOf": [
        {
          "$ref": "#/definitions/FirstInAnyOf"
        },
        {
          "$ref": "#/definitions/SecondInAnyOf"
        },
        {
          "properties": {
            "third": {
              "type": "string"
            }
          },
          "required": [
            "third"
          ]
        }
      ]
    }
  }
}
```

itaru2622:
@koxudaxi I want the following classes  with names from current test case.

- all classes are apeared and kept  the same names as the jsonschema.
- additional class has some suffix in its name if needed.

```python

class First(BaseModel): #  described in jsonschema, -> it should be apeared and kept its name
    first: str
class Second(BaseModel): #  described in jsonschema, -> it should be apeared and kept its name
    second: str
class TargetItem(BaseModel): # additional class to describe 'Target'
    third: str
class Target1(First, Target): # additional class
    pass
class Target2(Second, Target): # additional class
    pass
class Target(BaseModel):  # described in jsonschema,  but not appeared in  current PR's expected result.
# I want this class to be generated ;-)
    __root__: Union[Target1,Target2]
```

itaru2622:
@koxudaxi I checked call stacks for both of parse_all_of and parse_any_of, and found that:

1) in allOf case, the name of 'TargetItem'  for the class having  the 'third' member  is assigned  at below
https://github.com/koxudaxi/datamodel-code-generator/blob/afafa50c4ddd17e00a06e382c0a7638ce311e3a6/datamodel_code_generator/parser/jsonschema.py#L838-L843
   in parse_object via parse_item <- parse_list_item <- parse_any_of.
   
2) parse_item function cares most/all of types, like generic kids parser.
    this function  is also involved when toplevel is oneOf(parse_one_of), but NOT  if toplevel is allOf case.

so, it maybe better to refactor parse_all_of function to involve parse_item function.

special thanks to '@q.t' of https://github.com/zestyping/q for easy tracing :-)

koxudaxi:
@itaru2622 
Thank you for investigating the issue.
I fixed the code. And the naming rule is improved.
Could you please review it again?

itaru2622:
@koxudaxi thank you for your quick and hard work.

this fix works fine.

koxudaxi:
@itaru2622 
Thank you very much!!

Issue #1322, opened 2023-05-16
OpenAPI Scopes Don't Work With $ref


zach-hamm:
**Describe the bug**
There's an issue with generating models from the `parameters` and `paths` sections when using the `$ref` keyword. I tried to use a minimal example to get the point across, but let me know if you have additional questions. For now, I've been getting around this by using the swagger CLI to bundle my spec file and generating models off of that, but it's not the ideal approach as now I need all of my developers to download the CLI.

Note that the bug appears to be with the `paths.*` references. All references beyond that work as expected. For example, if I have a reference at `paths.<>.get.parameters.*` that works.

Note that I'm using 0.19.0

```
datamodel-codegen --version
0.19.0
```

---

**To Reproduce**

Example Schema
```yaml
openapi: 3.1.0
info:
  title: My API
  description: This is my API
  version: 3.0.0
servers:
  - url: /api
paths:
  /my/new/endpoint/{myParam}:
    $ref: '#/components/pathItems/MyNewEndpoint'
components:
  schemas:
    MyNewEndpointResponse:
      type: object
      properties:
        myProp:
          type: string
    MyNewEndpointName:
      type: string
  pathItems:
    MyNewEndpoint:
      get:
        description: "Gets stuff"
        operationId: myNewEndpoint
        tags:
          - MyEndpoints
        parameters:
          - name: myParam
            required: true
            in: path
            schema:
              type: string
        requestBody:
          content:
            application/json:
              schema:
                properties:
                  data:
                    properties:
                      name:
                        $ref: '#/components/schemas/MyNewEndpointName'
                    required:
                      - name
                type: object
          required: true
        responses:
          '200':
            description: OK
            content:
              application/json:
                schema:
                  $ref: '#/components/schemas/MyNewEndpointResponse'
tags:
  - description: My endpoints.
    name: MyEndpoints
```

Commandline Used:
```
datamodel-codegen --input ref.yaml --output ref.py --openapi-scopes schemas parameters paths
```

Resulting ref.py file (only schema objects are created):
```python
# generated by datamodel-codegen:
#   filename:  ref.yaml
#   timestamp: 2023-05-16T18:55:00+00:00

from __future__ import annotations

from typing import Optional

from pydantic import BaseModel


class MyNewEndpointResponse(BaseModel):
    myProp: Optional[str] = None


class MyNewEndpointName(BaseModel):
    __root__: str
```

[...]

PR #1323, opened 2023-05-16
Fix $ref With Path Items

zach-hamm:
I opened up an [issue for a bug with $ref on path items](https://github.com/koxudaxi/datamodel-code-generator/issues/1322). This PR applies the changes mentioned there to hopefully resolve the issue.


koxudaxi:
@zach-hamm 
Thank you very much!!

PR #1346, opened 2023-06-02
Fix the custom file header

Fokko:
Looks like the argument is missing :)


datamodel_code_generator/__init__.py
```diff
@@ -500,15 +500,18 @@ def get_header_and_first_line(csv_file: IO[str]) -> Dict[str, Any]:
...
         if path is None:
             file = None
         else:
             if not path.parent.exists():
                 path.parent.mkdir(parents=True)
             file = path.open('wt', encoding=encoding)
 
+        if custom_file_header and os.path.exists(custom_file_header):
+            with open(custom_file_header) as header_file:
+                custom_file_header = ''.join(header_file.readlines())
```
Fokko: @koxudaxi I took the liberty and added the option to read the header from file as well. We need to add the ASF header to each of the file in Iceberg: https://github.com/apache/iceberg/pull/7751/files Passing in a multi line string as an argument is a bit of a pain, so I figured this would be a nice addition.

datamodel_code_generator/__init__.py
```diff
@@ -500,15 +500,18 @@ def get_header_and_first_line(csv_file: IO[str]) -> Dict[str, Any]:
         header += f'\n#   version:   {get_version()}'
 
     file: Optional[IO[Any]]
-    for path, body_and_filename in modules.items():
-        body, filename = body_and_filename
+    for path, (body, filename) in modules.items():
```
koxudaxi, line 507: Oh, I didn't realize the way to unpack the tuple in the `for` statement.
Thank you!!

Fokko: I thought that was a nice addition, any time! 👍🏻

datamodel_code_generator/__main__.py
```diff
@@ -433,6 +433,15 @@ def sig_int_handler(_: int, __: Any) -> None:  # pragma: no cover
 arg_parser.add_argument(
     '--custom-file-header', help='Custom file header', type=str, default=None
 )
+
+arg_parser.add_argument(
+    '--custom-file-header-path',
+    help='Custom file header file path',
+    default=None,
+    type=str,
+)
```
koxudaxi, line 442: @Fokko 
I add a new option to avoid confusion.

Fokko: That works for me as well :) I was considering that as well

koxudaxi:
@Fokko 
Thank you very much!!

PR #1350, opened 2023-06-04
Added support for model wise base classes

senesh-deshan:
Added support for model wise base classes with `customBasePath` keyword in `JsonSchemaObject`.
It allows user to specify a custom base class per each generated model where as existing `base_class` argument is dedicated for specifying a single base class for all the models generated.


koxudaxi:
@senesh-deshan 
Thank you very much!!

PR #1374, opened 2023-06-20
pyproject: correct poetry-core package name

ConnorBaker:
I was trying to package this application with Nix and ran into a problem: instead of `poetry-core`, the `pyproject.toml` has `poetry_core`. The name of the package is `poetry-core`: https://github.com/python-poetry/poetry-core/blob/5e85c045b12c6150405ad8575b54f4e6d077aed9/pyproject.toml#L2.


koxudaxi:
@ConnorBaker 

PR #1448, opened 2023-07-24
main: fix configuration model validation after CLI args merge

### LINKED ISSUE #1435: Wrong parent class for pydantic V2 root models
**Describe the bug**
Generator uses `pydantic.BaseModel` as parent class for root model instead of `pydantic.RootModel`

Example schema (`custom_id.yaml`):
```yaml
openapi: 3.0.0
components:
  schemas:
    CustomId:
      description: My custom ID
      type: string
      format: uuid
```
Used commandline:
```
$ datamodel-codegen --input custom_id.yaml --output-model-type pydantic_v2.BaseModel --output model.py
```
Contents of `model.py`:
```python

from uuid import UUID

class CustomId(BaseModel):
    root: UUID = Field(..., description='My custom ID')
```

**Expected behavior**
```python

from uuid import UUID

from pydantic import RootModel, Field


class CustomId(RootModel):
    root: UUID = Field(..., description='My custom ID')
```

**Version:**
 - Python version: 3.10.8
 - datamodel-code-generator version: 0.21.1

airwoodix:
The bug reported in #1435 seems to originate in the `Config._validate_base_class` method: in Pydantic v2, `@model_validator(mode="after")` decorates an instance method instead of a class method that receives the raw values (the case for `mode="before"`). Therefore, except when called to alter `set_args` in `Config.merge_args`, the `values` argument is of type `Config` and `"base_class" not in values` is always true.

This MR is an attempt at fixing the issue, *for Pydantic v2 only* for the moment.

TODOS:
- [x] also fix the issue with Pydantic 1.x
- [x] add regression test for #1435.

Closes #1435.


koxudaxi:
> @model_validator(mode="after") decorates an instance method instead of a class method that receives the raw values (the case for mode="before"). 

@airwoodix 
I didn't know the changes.

Issue #1461, opened 2023-07-28
OpenAPI Parameters ignore base-class option


piercsi:
When using OpenAPIScope.Parameters scope, the resulting class ignores the base-class parameter
```
import json
from pathlib import Path
from tempfile import TemporaryDirectory

from datamodel_code_generator import generate, InputFileType, OpenAPIScope

open_api = {
    "openapi": "3.0.1",
    "paths": {
        "/Endpoint": {
            "get": {
                "parameters": [
                    {
                        "name": "limit",
                        "in": "query",
                        "schema": {
                            "type": "integer",
                            "format": "int32"
                        }
                    }
                ],
            }
        }
    }
}

with TemporaryDirectory() as output_dir:
    output = Path(output_dir) / "model.py"
    generate(
        json.dumps(open_api),
        input_file_type=InputFileType.OpenAPI,
        openapi_scopes=[OpenAPIScope.Paths, OpenAPIScope.Parameters],
        output=output,
        base_class='Models.MyModel',
        disable_timestamp=True
    )
    print(output.read_text())
```

**Expected behavior**
```
# generated by datamodel-codegen:
#   filename:  <stdin>

from __future__ import annotations

from typing import Optional

from Models import MyModel


class EndpointGetParametersQuery(MyModel):
    limit: Optional[int] = None
```

**Actual behavior**

```
# generated by datamodel-codegen:
#   filename:  <stdin>

from __future__ import annotations

from typing import Optional

from pydantic import BaseModel


class EndpointGetParametersQuery(BaseModel):
    limit: Optional[int] = None
```

**Version:**
 - OS: Windows 10
 - Python version: 3.11.4
 - datamodel-code-generator version: 0.21.2

**Additional context**

If this is by design please feel free to close.

piercsi:
My first ever PR. Go easy :)

koxudaxi:
@piercsi 
Thank you so much!!

PR #1462, opened 2023-07-28
Fix issue #1461: Open API parameter models use base class if set

piercsi:
As per #1461 this sets the base class for Open API parameter scopes to the base class supplied.


koxudaxi:
@piercsi 

PR #1485, opened 2023-08-09
switch away from deprecated pydantic method

iliakur:
I'm starting with just one method conversion to get initial feedback and check tests, but there's also one lurking in `types.py` which I'll convert later in the same PR.


koxudaxi:
Thank you for creating PR.
Could you please add the condition of `PYDANIC_V2` ?
Because, we should support v1 in runtime yet.
https://github.com/koxudaxi/datamodel-code-generator/blob/24bb074eda4354c254fc96c81d6242965af817a1/datamodel_code_generator/parser/jsonschema.py#L164-L175

koxudaxi:
@iliakur 
I just added `def model_rebuild(cls...` to DataType :wink: 

Thank you very much!!

iliakur:
@koxudaxi Thanks for the final tweaks! qq: when do you think this will be released?

koxudaxi:
@iliakur 
You can install the version from pypi via pip.
I'm sorry I didn't announce the released 
https://github.com/koxudaxi/datamodel-code-generator/releases/tag/0.21.4

iliakur:
oh, this is great, thanks a bunch for such a quick release!!

PR #1498, opened 2023-08-19
Fix default for annotated field in pydantic v2

i404788:
In pydantic v2 "assign-style" defaults don't work well with Annotated fields, you need to put them into the `Field` constructor. This handles that & also implements `use_default_kwarg` for the latest version.


koxudaxi:
@i404788 
The issue is fixed in the latest version.
Thank you very much!!

Issue #1278, opened 2023-04-25
Support for `msgspec.Struct` types as `--output-model-type`


jcrist:
Hello! First off - thanks for all your work on `datamodel-code-generator`, it's an impressive and useful piece of software.

I maintain [msgspsec](https://github.com/jcrist/msgspec), a fast JSON serialization and validation library, which offers similar (but much faster) functionality as Pydantic. We've had [a request](https://github.com/jcrist/msgspec/issues/390) to be able to generate `msgspec.Struct` types from existing json-schema definitions. I *think* the easiest way to do this would be to add `msgspec` integration into `datamodel-code-generator` as an `--output-model-type`. Is this a feature you'd be interested in supporting?

I'd be happy to help/do the work, although if I make the PR myself I'll likely need some guidance. `msgspec.Struct` types are closer to `dataclasses` than `pydantic.BaseModel` types, given you support both of them I'd *hope* many of the code paths could already be reused and adding an additional model type wouldn't be too onerous.

## Funding
* You can sponsor this specific effort via a [Polar.sh](https://polar.sh) pledge below
* We receive the pledge once the issue is completed & verified

<a href="https://polar.sh/koxudaxi/datamodel-code-generator/issues/1278">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/koxudaxi/datamodel-code-generator/issues/1278/pledge.svg?darkmode=1">
  <img alt="Fund with Polar" src="https://polar.sh/api/github/koxudaxi/datamodel-code-generator/issues/1278/pledge.svg">
</picture>
</a>

koxudaxi:
@jcrist 
Thank you for creating the issue.
I learned about your project for the first time, and it's cool :)

OK, I'm happy if you create the PR and I will help you.

We need to write some code for constraints.
https://jcristharif.com/msgspec/constraints.html#constraints

koxudaxi:
@jcrist 
The latest version supports `msgspec.Struct`
Thank you very much!!

jcrist:
Thanks!

PR #1551, opened 2023-09-16
Support msgspec output

indrat:
Adds basic support for `msgspec.Struct` as an output format which possibly addresses #1278 

I have not used this in anger yet, only to transform some of the examples so I cannot say whether it is completely correct but I figured it was a good idea to push it up for review. I basically followed the bouncing ball of changes from the previous two PRs to add `dataclass` and `TypedDict` support.

Things that I haven't figured out:
## Class args
Class args to Struct for things like `kw_only`, `forbid_unknown_fields`, or `omit_defaults`, i.e
```python
class Pet(Struct, kw_only=True, omit_defaults=True, forbid_unknown_fields=True):
    id: int
    name: str
    tag: Optional[str] = None
```
this is probably relevant for the dataclasses output as well for passing args to the `@dataclass` decorator

## discriminated unions
Somewhat relatedly - discriminated unions. msgspec uses class args, `tag` and `tag_field`, to denote when Structs participate in a tagged Union whereas the pydantic output includes that information on the Field. Not sure how to solve this one. Ideally output similar to the ClassVar approach in https://github.com/jcrist/msgspec/issues/338#issuecomment-1453701495

pydantic output
```python
# pydantic output of jsonschema discriminator_literals.json
class Type1(BaseModel):
    type_: Literal['a'] = Field('a', title='Type ')

class Type2(BaseModel):
    type_: Literal['b'] = Field('b', title='Type ')

class Response(BaseModel):
    inner: Union[Type1, Type2] = Field(..., discriminator='type_', title='Inner')
``` 

with this branch msgspec output has no way to add the `tag` or `tag_field` class args.
```python
# msgspec output of jsonschema discriminator_literals.json
class Type1(Struct):
    type_: Annotated[Literal['a'], Meta(title='Type ')] = 'a'

class Type2(Struct):
    type_: Annotated[Literal['b'], Meta(title='Type ')] = 'b'

class Response(Struct):
    inner: Annotated[Union[Type1, Type2], Meta(title='Inner')]
```

Possible approach using `ClassVar` from https://github.com/jcrist/msgspec/issues/338#issuecomment-1453701495 requires being able to add class args like `tag` / `tag_field`.
```python
class Base(msgspec.Struct, tag_field='type_'):
    type_: ClassVar[EventType]

class Type1(Struct, tag='a',):
    type_: Annotated[Literal['a'], Meta(title='Type ')] = 'a'

class Type2(Struct, tag='b'):
    type_: Annotated[Literal['b'], Meta(title='Type ')] = 'b'

class Response(Struct):
    inner: Annotated[Union[Type1, Type2], Meta(title='Inner')]
```


datamodel_code_generator/__main__.py
```diff
@@ -283,6 +283,9 @@ def merge_args(self, args: Namespace) -> None:
             if getattr(args, f) is not None
         }
 
+        if set_args.get('output_model_type') == DataModelType.MsgspecStruct.value:
+            set_args['use_annotated'] = True
```
koxudaxi, line 287: @indrat 
Doesn't msgspec use `Field`? and is it always annotated?
You define some logic for `Field`. But, the unit tests don't cover the parts.
We should change the coverage to 100% before merging.

koxudaxi:
@indrat 
Thank you for creating the great PR.
I fix the isort problem of unittest.

The unittests don't cover all lines.
I left one comment.
Could you please check it?

datamodel_code_generator/__main__.py
indrat: I've added some output tests for when a `msgspec.field` is generated and tried to boost the coverage for the const/alias fields as well.

Regarding forcing`use_annotated` I believe the constraints such as `gt`, `lt`, etc must be attached using `Annotated[..., Meta(gt=..., lt=...)]` etc. So aiming for the least surprise I felt it made sense to always force `use_annotated` on otherwise you won't get the constraints or description metadata generated.

[`msgspec.field`](https://jcristharif.com/msgspec/api.html#msgspec.field) only accepts `default`,`default_factory`, and `name` so it's a little different to `pydantic.Field`

koxudaxi:
Thank you very much!!

Issue #1491, opened 2023-08-10
New type error for Pydantic v2 models in 0.21.4 release


willarmiros:
**Describe the bug**
When upgrading from 0.21.3 -> 0.21.4, my Pydantic v2 models were generated in a slightly different way that involved using a `RootModel`, which in turn triggered a type error from MyPy.

**To Reproduce**

Example schema:
```json

{
    "openapi": "3.1.0",
    "info": {
        "title": "FastAPI",
        "version": "0.1.0"
    },
    "components": {
        "schemas": {
            "MySchema": {
                "properties": {
                    "name": {
                        "type": "string",
                        "title": "Name"
                    },
                    "uuid": {
                        "type": "string",
                        "title": "Uuid"
                    },
                    "external_uri": {
                        "anyOf": [
                            {
                                "type": "string"
                            },
                            {
                                "type": "null"
                            }
                        ],
                        "title": "External Uri"
                    },
                    "description": {
                        "anyOf": [
                            {
                                "type": "string",
                                "maxLength": 256
                            },
                            {
                                "type": "null"
                            }
                        ],
                        "title": "Description"
                    }
                },
                "type": "object",
                "required": [
                    "name",
                    "uuid"
                ],
                "title": "MySchema"
            }
        },
        "securitySchemes": {
            "HTTPBearer": {
                "type": "http",
                "scheme": "bearer"
            }
        }
    }
}

```

Used commandline:
```
$ poetry run datamodel-codegen --input repro-openapi.json --input-file-type openapi --use-annotated --output-model-type "pydantic_v2.BaseModel" --base-class "pydantic.BaseModel" --output model.py
```

MyPy commands (using the latest MyPy 1.5.0, but this also happened with MyPy 1.4.1):

```
poetry run mypy --ignore-missing-imports --strict --check-untyped-defs .
````

And the error output:

```
model.py:13: error: Missing type parameters for generic type "RootModel"  [type-arg]
````

**Expected behavior**
Expected behavior is that no MyPy errors are raised as was the case before.

**Version:**
 - OS: Mac OSX
 - Python version: Python 3.11.3
 - datamodel-code-generator version: 0.21.4

**Additional context**

 I was able to resolve this issue by manually adding the type parameter to the RootModel base class, like so:

```python
# Original generated by datamodel-codegen
class DescriptionItem(RootModel):
    root: Annotated[str, Field(max_length=256)]

# My fix
class DescriptionItem(RootModel[str]):
[...]

PR #1560, opened 2023-09-21
fix: add type parameters to generic RootModel

tcrasset:
Closes: https://github.com/koxudaxi/datamodel-code-generator/issues/1491

and by extension it fixes mypy's error:

`Missing type parameters for generic type 'RootModel'`

It's not clear from the [pydantic documentation](https://docs.pydantic.dev/dev-v2/usage/models/#rootmodel-and-custom-root-types) whether they force the user to use declare the type twice. They have one example with `Pets(RootModel[List[str]])` and one with  `Pets(RootModel)`. Though `mypy` seems to be in favor of the double type hint style.

### DIFF
--- a/datamodel_code_generator/model/template/pydantic_v2/RootModel.jinja2
+++ b/datamodel_code_generator/model/template/pydantic_v2/RootModel.jinja2
@@ -1,7 +1,16 @@
+{%- macro get_type_hint(_fields) -%}
+{%- if _fields -%}
+{#There will only ever be a single field for RootModel#}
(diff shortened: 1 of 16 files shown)


tcrasset:
Hi @koxudaxi, the CI seemed to fail only due to the code coverage report not succeeding to upload. The tests seem to pass.
Please review :)

koxudaxi:
@tcrasset 
I didn't know RootModel is a generic type.
Thank you very much!! :smile:

Issue #1607, opened 2023-10-06
msgspec output of a directory input where a schema only contains a root model is missing imports


indrat:
**Describe the bug**
Similar to #1495 when using a directory input with a annotated constrainted `RootModel` where the schema ends up generating solely `RootModel`s in the generated python, the module is missing the required imports.

**To Reproduce**

```
$ tree schemas
schemas
├── common.yml
└── test.yml

$ datamodel-codegen --input schemas --input-file-type jsonschema --output src --disable-timestamp --enable-version-header --output-model-type msgspec.Struct

$ tree src
src
├── __init__.py
├── common.py
└── test.py

# Note the missing `from msgspec import Meta`
$ cat common.py
# generated by datamodel-codegen:
#   filename:  test.yml
#   version:   0.22.0

from __future__ import annotations

from typing import Any

from typing_extensions import Annotated

Model = Any


Ulid = Annotated[str, Meta(pattern='[0-9ABCDEFGHJKMNPQRSTVWXYZ]{26,26}')]
```

Example schema:
- schemas/common.yml
```yaml
---
$schema: https://json-schema.org/draft/2020-12/schema
$id: common.yml
definitions:
  ulid:
    type: string
    pattern: '[0-9ABCDEFGHJKMNPQRSTVWXYZ]{26,26}'
```
- schemas/test.yml
```yaml
---
$schema: https://json-schema.org/draft/2020-12/schema
$id: test.yml
title: test
required:
  - uid
properties:
  uid:
    description: ulid of this object
    $ref: ./common.yml#/definitions/ulid
```

Used commandline:
```
$ datamodel-codegen --input schemas --input-file-type jsonschema --output src --disable-timestamp --enable-version-header --output-model-type msgspec.Struct
```

**Expected behavior**
Generated modules that only root models that are annotated with `Meta` constraints import the required classes.

eg:
```python
from msgspec import Meta
```

**Version:**
 - OS: Linux, debian 12.
 - Python version: 3.11
 - datamodel-code-generator version: 0.22.0

indrat:
Investigating this it would appear that it maybe because the `root_model.RootModel` has no concept of adding to the imports so when a python module is generated that only contains root models the requisite imports aren't added?

Although there is obviously some interaction with the `msgspec.Struct` model as the root model has been able to "acquire" the `Meta` constraint however I've no idea how that happens or how to affect the import list for a root model.

koxudaxi:
[...]

PR #1611, opened 2023-10-07
Fix msgspec root import

koxudaxi:
Closes: https://github.com/koxudaxi/datamodel-code-generator/issues/1607


koxudaxi:
@indrat 
I have moved the imports method to decorator. I apply it to RootModel too.

Could you please review it?

(Maybe, we should check another output model type too? :thinking: )

indrat:
lgtm 👍 thank you for your amazing work :)

PR #1618, opened 2023-10-13
Use date/time/datetime Python types

### LINKED ISSUE #1615: Format is ignored for output model type dataclass
**Describe the bug**

The format of a string property is not considered for the output model type `dataclasses.dataclass`.

**To Reproduce**

Example schema:

```yaml
openapi: 3.0.3
info:
  title: Bug report
  version: 0.0.1
paths:
  /api:
    get:
      responses:
        '200':
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/Model'
          description: ''
components:
  schemas:
    Model:
      type: object
      properties:
        updated:
          type: string
          format: date-time
        start:
          type: string
          format: date
```

Used commandline:
```
$ datamodel-codegen --input openapi.yaml --output models.py  --output-model-type dataclasses.dataclass --input-file-type openapi
```

**Expected behavior**

The output file should contain the correct types for the date time/date/time etc. like with the default output type:

```python
@dataclass
class Model:
    updated: Optional[datetime] = None
    start: Optional[date] = None
```

Currently, the types are just `str`.

**Version:**
 - OS: Windows/Linux
 - Python version: 3.10
 - datamodel-code-generator version: 0.22.1

twwd:
according to the OpenAPI property type for all output model types

Fixes #1615

### DIFF
--- a/datamodel_code_generator/model/types.py
+++ b/datamodel_code_generator/model/types.py
@@ -1,7 +1,7 @@
 from typing import Any, Dict, Optional, Sequence, Type
 
 from datamodel_code_generator import PythonVersion
-from datamodel_code_generator.imports import IMPORT_ANY, IMPORT_DECIMAL
+from datamodel_code_generator.imports import IMPORT_ANY, IMPORT_DECIMAL, IMPORT_DATETIME, IMPORT_DATE, IMPORT_TIME
 from datamodel_code_generator.types import DataType, StrictTypes, Types
 from datamodel_code_generator.types import DataTypeManager as _DataTypeManager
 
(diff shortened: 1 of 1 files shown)


twwd:
It might be better to set the type to `datetime.time` etc. instead of just `time` etc. to prevent naming clashes.

koxudaxi:
@twwd 
In pydantic, parsing allows normal strings to be cast to `datetime.datetime`, but this is not done in dataclass or typedDict. Considering that OpenAPI is used as an interface definition with outside Python, wouldn't this change no longer match the use cases for `dataclass` or `TypedDict`? There will probably be a need to cast from string to `datetime.datetime` before assigning it to a dataclass.

twwd:
I parse the server JSON response into a `dataclass` and handle the type conversion but I see that this is not the general use-case.
I patch the method as a workaround (if somebody else has a similar use-case)

```python
with mock.patch('datamodel_code_generator.model.types.type_map_factory') as type_map_factory_mock:
   type_map_factory_mock.side_effect = adjusted_type_map_factory
```

PR #1684, opened 2023-11-12
Relative imports: Import class instead of module

### LINKED ISSUE #1683: Relative imports paths when generating multiple models
When generating schemas with a directory of JSON-Schemas as `--input` the relative import paths will be problematic. I had a complex use-case where pydantic threw an error when attempting to import such a generated model. But maybe I'm missing something.

So, e.g. if you have a structure like
```
- src
    - pkg1
        - a.py
    - pkg2
        - b.py
```
a module would look like this:
```python
from ..pkg2 import b

class A(BaseModel):
    bb: b.B | None = None
```

Which is fine in principle. But when attempting to import e.g. class `A` which references to an enum, pydantic throws an error similar to `B has no attribute B`. So this might be more an issue of pydantic. I would suggest to change the relative import to import the model directly (or at least have a flag for this) anyway.

```python
from ..pkg2.b import B

class A(BaseModel):
    bb: B | None = None
```

lord-haffi:
This would resolve #1683.
This PR is just a suggestion. I monkey-patched the function in my code this way to get my use-case to work so I wanted to share it with you. But maybe another approach e.g. using a flag to support both "import-styles" could be better, idk.


lord-haffi:
@koxudaxi Apparently, this would break many Unittests due to unexpected outputs. Perhaps, using an optional flag would be better.

koxudaxi:
@lord-haffi 
I'm sorry for my late response.
I agree with you. I will check it when I get the time.

lord-haffi:
Superseded by #1983

PR #1691, opened 2023-11-14
Fix error when JsonSchemaObject is bool with Pydantic v2

sbussard-vareto:
I don't know what the ramifications of this change are, but I know this change seems to skip the error when data classes from json schema


sbussard-vareto:
Wow thanks so much for using this workaround to make an actual fix! Such a fast turnaround!

mattholden:
Any idea when this will be available as a patch release?

Issue #1643, opened 2023-10-26
Add ruff as optional formatter in runtime


koxudaxi:
[ruff](https://github.com/astral-sh/ruff) supports `format` command.
We should add the feature to select `ruff` as formatter in code generation instead of `black`.

LordFckHelmchen:
@koxudaxi This can be closed as implemented, right?

Issue #621, opened 2021-11-19
Ability to add custom formatters


adaamz:
**Is your feature request related to a problem? Please describe.**
Currently I'm trying to workaround some problem described before, so I created some really strange template with imports but generator also adds some imports which are not used anymore. So I would like to remove those unused imports with for example `autoflake` (because isort/black cannot do this atm and looks like it will never be their ability).

**Describe the solution you'd like**
The most variable solution is probably ability to add custom formatters (or in general scripts to be run after generation is completed).
This could be configured easily in pyproject.toml like this
```toml
[tool.datamodel-codegen.formatter.autoflake]
executable = autoflake
args = --remove-unused-variables
```

**Describe alternatives you've considered**
Creating own makefile target for this - combined running generator and then in next round running autoflake

koxudaxi:
I added the features

PR #1733, opened 2023-11-23
Add custom formatters

denisart:
Hooks in other libs: https://the-guild.dev/graphql/codegen/docs/config-reference/lifecycle-hooks

it is beta version of custom code formatters. it is simple realization without lifecycle hooks. Only before code generation

See `docs/custom-formatters.md` for simple usage example. And see `test_format.py` for development examples.

What do you think about this idea?


koxudaxi:
@denisart 
We need the feature to use another formatter. We have the feature request. https://github.com/koxudaxi/datamodel-code-generator/issues/621

This PR provides a flexible way to pass the formatter.
But, can we give the preset formatter like ruff? https://github.com/koxudaxi/datamodel-code-generator/issues/1643
(I want to change the default formatter to ruff. Because black spend time in the process.)
image https://codspeed.io/koxudaxi/datamodel-code-generator/branches/denisart:add-custom-formatters


I expect the CLI arguments will be added in the future.
```
--preset-formatter-python [ruff, black-isort]
or 
--preset-formatter [ruff, black-isort, gofmt]
```

datamodel_code_generator/format.py
```diff
@@ -167,12 +170,49 @@ def __init__(
...
     def format_code(
         self,
         code: str,
     ) -> str:
         code = self.apply_isort(code)
         code = self.apply_black(code)
+
+        for formatter in self.custom_formatters:
+            code = formatter.apply(code)
+
```
koxudaxi, line 215: Should the method apply all formatter?
Someone may want to replace black/isort formatters with others like ruff.

denisart: > Should the method apply all formatter?

I also agree. I think that in the future we need to add popular custom formatters and freedom of choice.

I think that this method should apply the list of formatters specified by the user like this
```bash
datamodel-codegen ... \
  --formatters \
  "datamodel_code_generator.ruff,user_formatter_module"
```

denisart:
> But, can we give the preset formatter like ruff?

I agree with the ruff as default (if user not specified formatters). Do you want to work on this PR or as part of a new PR?

datamodel_code_generator/format.py
denisart: > ... apply the list of formatters specified by the user ...

or default formatter (ruff) or disable format (special flag)

koxudaxi: OK, What do you think about the option?
```
--default-formatter  [black-isort(default), ruff, none]
```` 
`none` maybe not clearly?
Or
```
--default-formatter  [black-isort(default), ruff]
and 
--disable-default-formatter 
````

koxudaxi:
@denisart 
> I agree with the ruff as default (if user not specified formatters). Do you want to work on this PR or as part of a new PR?

Thank you for asking.
I don't expect to set `ruff` as default soon. Because I guess the changes will affect many users.
But, I prefer to introduce `ruff` option as preset.

This project is used by many projects. I want to avoid breaking changes in the short term. (including adding dependency )

datamodel_code_generator/format.py
denisart: The second option seems clearer to me.

Then we have 4 args:

`--default-formatter`: choice one (or a few ?) default formatters;
`--disable-default-formatter`: disable default formatters
`--custom-formatters`: user formatters
`--custom-formatters-kwargs`: json file with params for user formatters

koxudaxi: Looks good.
We have discussed generating non-python code by the CLI.
We should accept another language formatter or turn off the default formatter.
https://github.com/koxudaxi/datamodel-code-generator/discussions/1657

denisart: > generating non-python code by the CLI

It is good idea! I have a few questions about this.
I will take a look and answer in the discussion.

denisart:
> I want to avoid breaking changes in the short term.

I agree with you.

datamodel_code_generator/format.py
denisart: @koxudaxi 

I answer here. I agree with [#1657](https://github.com/koxudaxi/datamodel-code-generator/discussions/1657).
I'm thinking about a more universal approach myself. I have some thoughts about realization details.
I'll try to move in this direction in the next commits.

denisart: Maybe we can create a discussion where we discuss "road map" for `datamodel-code-generator` version `1.0.0`?

koxudaxi: I just created it.
https://github.com/koxudaxi/datamodel-code-generator/discussions/1736

denisart:
@koxudaxi 
Do we have any undiscussed issues about this PR?

koxudaxi (approved):
LGTM.

PR #1746, opened 2023-11-27
Add formatter module

denisart:
1. Added a new formatter module (now is not using in main code).

Example:
```python
from datamodel_code_generator.formatter.base import BaseCodeFormatter

class CustomHeaderCodeFormatter(BaseCodeFormatter):
        formatter_name: ClassVar[str] = "custom"

        def __init__(self, formatter_kwargs: Dict[str, Any]) -> None:
            super().__init__(formatter_kwargs=formatter_kwargs)
            default_header = "my header"
            self.header: str = self.formatter_kwargs.get("header", default_header)

        def apply(self, code: str) -> str:
            return f'# {self.header}\\n{code}'
        
formatter_kwargs = {"header": "formatted with CustomHeaderCodeFormatter"}
formatter = CustomHeaderCodeFormatter(formatter_kwargs)

code = '''x = 1\ny = 2'''
print(formatter.apply(code))
"""
# formatted with CustomHeaderCodeFormatter
x = 1
y = 2
"""
```

2. Added isort, black and ruff (empty now) formatters based on `datamodel_code_generator.formatter.base.BaseCodeFormatter`.

3. New formatters runner (based on `CodeFormatter` from `datamodel_code_generator.format`). Usage example:

```python
runner = CodeFormattersRunner(
    default_formatter=['black', 'isort'],
    disable_default_formatter=False,
    custom_formatters=["my_package.my_sub_package.CustomHeaderCodeFormatter"],
    custom_formatters_kwargs={
        'custom': {
            'header': 'formatted with CustomHeaderCodeFormatter',
        }
    }
)

runner.format_code("...")
# running black
# running isort
# running custom
```


koxudaxi:
@denisart could you please rebase the branch?
I just updated the poetry version and merged it to main.

koxudaxi:
@denisart the best is that you add the write permission of the PR to me

datamodel_code_generator/formatter/base.py
```diff
@@ -0,0 +1,184 @@
...
+
+        custom formatter
+        >>> custom_formatter_import = "my_package.my_sub_package.FormatterName"
+        this is equivalent to code
+        >>> from my_package.my_sub_package import FormatterName
+
+    """
+
+    import_ = Import.from_full_path(custom_formatter_import)
+    imported_module_ = import_module(import_.from_)
```
koxudaxi, line 73: I didn't think the uses case :sweat_smile:

datamodel_code_generator/formatter/base.py
```diff
@@ -0,0 +1,184 @@
...
+
+    def _check_default_formatters(
+        self,
+        default_formatters: Optional[List[str]],
+    ) -> List[BaseCodeFormatter]:
+        if self.disable_default_formatter is True:
+            return []
+
+        if default_formatters is None:
+            return self._load_formatters(self._default_formatters)
```
koxudaxi, line 153: Why does the method expect the path string? 
I prefer `custom_formatter: BaseCodeFormatter`  to raw str.
We can use type-checking.

If we want to pass the custom formatted class from CLI, how about loading the custom formatted class from str only for that?
I feel that passing raw `str` is a bit confusing, since it is also intended to be used as a module by the user.

koxudaxi (commented):
@denisart 
Thank you for the great PR.
I left a few comments.

datamodel_code_generator/formatter/base.py
denisart: I also had the same reaction. But the `Import` class is very suitable in this case.

datamodel_code_generator/formatter/base.py
denisart: It seems I didn't fully understand your proposal. Can you explain or give an example?

koxudaxi: @denisart 
I'm sorry I show you the code.
I expect the arguments.
```python
class CodeFormattersRunner:
    ...
    def __init__(
        self,
        disable_default_formatter: bool = False,
        default_formatter: Optional[List[BaseCodeFormatter]] = None,
        custom_formatters: Optional[List[BaseCodeFormatter]] = None,
```

if we give the external formatter to `CodeFormattersRunner`, we should get the `BaseCodeFormatter` by using 
 `load_code_formatter`
 
 ```python
custom_comatter = load_code_formatter("my_package.my_sub_package.FormatterName")
runner = CodeFormattersRunner(custom_formatters=[custom_comatter])
 ```
I don't think `CodeFormattersRunner` must know the full path string; `CodeFormattersRunner` only needs to handle the CodeFormatter class. What do you think?

denisart: @koxudaxi ,
I understood your idea. Do you propose to move the loading of formatters from `CodeFormattersRunner`? This is good from an architectural point of view.

Do you propose loading formatters by path in `generate` method from `datamodel_code_generator/__init__.py`?

koxudaxi: @denisart 
> I understood your idea. Do you propose to move the loading of formatters from CodeFormattersRunner? This is good from an architectural point of view.

Yes!!

> Do you propose loading formatters by path in generate method from datamodel_code_generator/__init__.py?

Yes, it is. Thank you for saying what I was trying to say.

denisart: Okay, I agree with you. I'll create a fix.

gaborbernat:
Seems stalled, closing we can reopen if you pick it up again.

PR #1783, opened 2023-12-21
Enable ability to use custom `JsonSchemaObject` in `JsonSchemaParser`

WilliamJamieson:
As described in #1781, I have a need to extend the `JsonSchemaParser` to handle an "extension" to JSON schema. In that discussion, I noted that I figured out method for achieving this, but it required duplication with slight alteration of code already present in the `JsonSchemaParser`.

This PR is a possible solution to avoiding that issue. It does this by factoring the direct call to the `JsonSchemaObject` out into a class variable which can be overridden to a subclass of the `JsonSchemaObject` by a sub classing the `JsonSchemaParser`. If there are better alternate solutions to this issue feel free to close this PR and suggest them in #1781.

### DIFF
--- a/datamodel_code_generator/parser/jsonschema.py
+++ b/datamodel_code_generator/parser/jsonschema.py
@@ -367,6 +367,7 @@ EXCLUDE_FIELD_KEYS = (
 @snooper_to_methods(max_variable_length=None)
 class JsonSchemaParser(Parser):
     SCHEMA_PATHS: ClassVar[List[str]] = ['#/definitions', '#/$defs']
+    SCHEMA_OBJECT_TYPE: ClassVar[Type[JsonSchemaObject]] = JsonSchemaObject
 
     def __init__(
         self,
@@ -686,7 +687,7 @@ class JsonSchemaParser(Parser):
                 # }
             else:
                 combined_schemas.append(
-                    JsonSchemaObject.parse_obj(
+                    self.SCHEMA_OBJECT_TYPE.parse_obj(
                         self._deep_merge(
                             base_object,
                             target_attribute.dict(exclude_unset=True, by_alias=True),
@@ -1632,7 +1633,7 @@ class JsonSchemaParser(Parser):
         raw: Dict[str, Any],
         path: List[str],
     ) -> None:
-        self.parse_obj(name, JsonSchemaObject.parse_obj(raw), path)
+        self.parse_obj(name, self.SCHEMA_OBJECT_TYPE.parse_obj(raw), path)
 
     def parse_obj(
         self,
@@ -1759,7 +1760,7 @@ class JsonSchemaParser(Parser):
                 # Some jsonschema docs include attribute self to have include version details
                 raw.pop('self', None)
                 # parse $id before parsing $ref
-                root_obj = JsonSchemaObject.parse_obj(raw)
+                root_obj = self.SCHEMA_OBJECT_TYPE.parse_obj(raw)
                 self.parse_id(root_obj, path_parts)
                 definitions: Optional[Dict[Any, Any]] = None
                 for schema_path, split_schema_path in self.schema_paths:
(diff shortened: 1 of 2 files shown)


koxudaxi:
@WilliamJamieson 
Should we add the change to OpenAPIParser too?

I think that is a good approach. Can you please add a brief unittest?

WilliamJamieson:
> Should we add the change to OpenAPIParser too?
Honestly, I have no clue if this would be useful for OpenAPI or not. Someone who has experience with it might be a better judge of that.

> Can you please add a brief unittest?
I will see what I can do. Please let me know if you have any thoughts for minimally testing this.

WilliamJamieson:
I made an extremely contrived example for this.

WilliamJamieson:
Darn it looks like it doesn't work for Pydantic v1. I'll have to think about this.

koxudaxi:
@WilliamJamieson 
pydantic v1's union behavior is not same as v2.  https://docs.pydantic.dev/1.10/usage/types/#unions

datamodel_code_generator/parser/jsonschema.py
```diff
@@ -367,6 +367,7 @@ def _get_type(type_: str, format__: Optional[str] = None) -> Types:
 @snooper_to_methods(max_variable_length=None)
 class JsonSchemaParser(Parser):
     SCHEMA_PATHS: ClassVar[List[str]] = ['#/definitions', '#/$defs']
+    SCHEMA_OBJECT_TYPE: ClassVar[Type[JsonSchemaObject]] = JsonSchemaObject
```
koxudaxi, line 370: @WilliamJamieson 
Why not just pass it through init and make it an instance variable?
This is not a criticism but a simple question.

WilliamJamieson: There is no particular reason. My application already required me to make a subclass of the parser itself to introduce some other behaviors that I need. So it makes no difference to me whether it's passed in through the `__init__` or a class variable.

I figured this type of change could drastically change the parser behavior. So forcing users to do it via a subclass made sense. L

WilliamJamieson:
> Darn it looks like it doesn't work for Pydantic v1. I'll have to think about this. Honestly, I have no clue why it doesn't work there as I have only ever used Pydantic v2.

I think the issue is my use of `model_post_init`. That does not appear to be documented in Pydantic v1. Maybe switching it to using `__post_init__` might work? I'll try to investigate this in the morning.

tests/parser/test_jsonschema.py
```diff
@@ -464,3 +467,72 @@ def test_no_additional_imports():
...
+    """
+    Contrived example to extend the JsonSchemaParser to support an alt_type, which
+    replaces the type if present.
+    """
+
+    class AltJsonSchemaObject(JsonSchemaObject):
+        properties: Optional[Dict[str, Union[AltJsonSchemaObject, bool]]] = None
+        alt_type: Optional[str] = None
+
+        def model_post_init(self, context: Any) -> None:
```
WilliamJamieson, line 526: > I think the issue is my use of `model_post_init`. That does not appear to be documented in Pydantic v1. Maybe switching it to using `__post_init__` might work? I'll try to investigate this in the morning.

This was not entirely correct, I can fix the failure in Pydantic 1.7+, via 
```suggestion
        def __init__(self, *args, **kwargs):
            super().__init__(*args, **kwargs)
```
However, I run into pydantic/pydantic#1985 on versions older than that. Adjusting the skip down to `< 1.7` does not prevent the failure in this case as it occurs during test collection not test running.

So its probably best to leave the test as is, and keep the skip.

koxudaxi: @WilliamJamieson 
Thank you for your research.
Perhaps this is not a problem with the code generator, but with how Pydantic is used.
It should be the responsibility of the creator of the custom schema model.

koxudaxi (approved):
LGTM

datamodel_code_generator/parser/jsonschema.py
koxudaxi: make sense!!

koxudaxi:
Thank you very much!!

WilliamJamieson:
@koxudaxi thanks so much for accepting my pull request.

PR #1826, opened 2024-01-30
Stop modifying behavior of yaml on load

### LINKED ISSUE #1825: Importing code generator changes behavior of pyyaml
When importing ``datamodel_code_generator``, the behaviour of ``pyyaml`` is changed after the import, datetime objects suddenly are deserialized from yaml as `str` whereas before they were deserialized as `datetime` objects.

### create clean venv
```bash
python3 -m venv blub
blub/bin/pip install datamodel-code-generator pyyaml
```

### Run this reproducer with the interpreter from the venv
```python
import yaml
import tempfile
import datetime

tmp=tempfile.NamedTemporaryFile(mode='w')

def write_read():
    tmp=tempfile.NamedTemporaryFile(mode='w')
    yaml.safe_dump({"datetime": datetime.datetime.now()}, tmp)
    dumped_reloaded = yaml.safe_load(open(tmp.name,"r"))
    print(dumped_reloaded)

print("before importing datamodel-code-generator:")
write_read()

import datamodel_code_generator
print("\nafter importing datamodel-code-generator:")
write_read()
```

**Actual output:** (time string + object may change depending on run time)
```
before importing datamodel-code-generator:
{'datetime': datetime.datetime(2024, 1, 30, 9, 51, 19, 612368)}

after importing datamodel-code-generator:
{'datetime': '2024-01-30 09:51:20.019991'}
```

**Expected output:**  (time string + object may change depending on run time):
```
before importing datamodel-code-generator:
{'datetime': datetime.datetime(2024, 1, 30, 9, 51, 19, 612368)}

after importing datamodel-code-generator:
{'datetime': datetime.datetime(2024, 1, 30, 9, 51, 20, 019991)}
```

The reason seems to be that the SafeLoader of ``yaml`` is modified globally by ``datamodel-code-generator``
https://github.com/koxudaxi/datamodel-code-generator/blob/acc6bf604b13626f22fc123d72ae08ff0a114155/datamodel_code_generator/util.py#L72

gluap:
Only modify a copy of ``SafeLoader`` to ensure that we don't change the behaviour of ``pyyaml`` for everyone.

Fixes #1825


gluap:
Note that CodeQL is happy now - the failing tests seem to be down to problems with the github workflow missing a token - likely unrelated to the content of the PR, as nothing in that direction has changed..

koxudaxi:
@gluap 
you are right. we should copy it.

> Note that CodeQL is happy now - the failing tests seem to be down to problems with the upload of coverage because the github workflow is missing a token to upload to codecov.io - likely unrelated to the content of the PR, as nothing in that direction has changed.

OK, I will fix it to merge the PR.

PR #1846, opened 2024-02-09
Added graphql to `--input-file-type` options

JoeHCQ1:
This should save future users a bit of guessing.


denisart:
Thanks! Can you adding the same word to `docs/index.md`?

PR #1879, opened 2024-03-08
Added support for passing pathlib.Path as a format to JSON schema

brandonsorensen:
I'm often in a situation where my models have attributes that refer logically to paths on the host machine. I wondered if there is any interest in incorporating the standard library's `pathlib.Path` class as a valid string format. I focused on JSON schema inputs, as they are my own primarily use case. As the changes are rather trivial, I drafted a quick implementation instead of opening a PR.


koxudaxi:
@barreeeiroo 
Looks good.
Could you please unittest for the feature?

PR #1911, opened 2024-04-10
Support http query parameters

kevin-lithic:
The goal of this is to be able to specify some set of query parameters that can be used on all calls when pulling schemas. This is particularly important because we want to be able to use GitHub to manage our schema versions and allow developers to generate their schemas from custom branches.


kevin-lithic:
@koxudaxi do I need to do something to get this to move thru the status checks?

Issue #1170, opened 2023-03-09
Split output in individual .py files


wabiloo:
**Describe the solution you'd like**
I would like an option to be able to output 1 .py file per model, instead of having them all in the same file. Ideally with then a  _init_.py file that imports them all.

**Describe alternatives you've considered**
Post-generation treatment of the file (somehow) to split it, but that looks complex

alexpovel:
Second this.

For example, the GitHub API responses surface is absolutely enormous and very fragmented (*repository* is one of the core concepts of course, but different endpoints return slightly different versions):

```console
$ python --version
Python 3.11.2
$ datamodel-codegen --version
0.21.4
$ datamodel-codegen --openapi-scope paths --url https://raw.githubusercontent.com/github/rest-api-description/v2.1.0/descriptions/api.github.com/dereferenced/api.github.com.deref.json > api.py
[warnings redacted]
$ wc -l api.py
94887 api.py
$ grep -P 'class Repository\d*\(BaseModel\)' api.py | wc -l
90
```

So we have `Repository` all the way through to `Repository89`. That is sadly too unwieldy. Having this many duplicates or nigh-duplicates would be fine if they'd be namespaced.

One workaround I am in the process of exploring is something along the lines of:

```console
jq '.paths["/orgs/{org}"]["get"]["responses"]["200"]["content"]["application/json"]["schema"]' api.github.com.2022-11-28.deref.json > org.json
```

For this, download the [REST OpenAPI description](https://github.com/github/rest-api-description/tree/c8142ac6c70bf74a638137deeecdae4b7fc8eb61/descriptions/api.github.com).

Now `org.json` contains a workable subset. If we mirror that to the local file system, replacing variables like `{org}` with, say `ORG`, we could get a workable solution:

```console
$ datamodel-codegen --input org.json --output src/github/api/orgs/ORG/__init__.py
```

Now, inside the `src` Python package, one can issue:

```python
from src.github.api.orgs.ORG import OrganizationFull as Organization
```

So a workaround can look like this (haven't implemented yet):

1. Fetch full Open API spec from GitHub (41 MB)
2. for all paths (view with `jq '.paths | keys' api.json`), extract its schema (if no `get` and `200` keys available, skip, I guess, if we're only querying)
3. save as `schema.json` to some new path mirroring the API one (which is nice as it makes it easy to match with the docs, and makes it google-able); make it filesystem- and more importantly, Python `import`-friendly (remove `{` etc.)
[...]

PR #1913, opened 2024-04-11
Support outputting all referenced http files to seperate files in folder

kevin-lithic:
Open to any feedback on this. Our goal here is to be able to parse an http URL based schema into separate files. This will allow a URL to be passed in an any new types it creates will be created in a new file

It looks like it is related to this open issue: https://github.com/koxudaxi/datamodel-code-generator/issues/1170


koxudaxi:
@kevin-lithic 
Could you please add unittests to improve coverage?

datamodel_code_generator/reference.py
```diff
@@ -555,7 +562,12 @@ def join_path(path: Sequence[str]) -> str:
             joined_path += '#'
         return joined_path
 
-    def add_ref(self, ref: str, resolved: bool = False) -> Reference:
+    def add_ref(
+        self,
+        ref: str,
+        resolved: bool = False,
+        http_folder_output: Optional[Path] = None,
```
koxudaxi, line 569: @kevin-lithic 
If `http_folder_output` is always specified, how about adding it to `__init__`?
Of course, if `http_folder_output` is added for the same instance and the case of `None` occurs, you can leave it as is.

kevin-lithic:
@koxudaxi thanks for the feedback, I wasnt sure if this was the right approach or not so just hacked something together to get the general idea across! Going to close this one and will open with a cleaned up version based on your input. Thanks again.

Issue #1955, opened 2024-05-11
Casts `default` values of type `number` (scientific notation) to `str`


maximilian-tech:
**Describe the bug**
Bug in parsing `number` in scientific notation

**To Reproduce**

Example schema:
```json
{
    "$schema": "http://json-schema.org/draft-07/schema#",
    "type": "object",
    "properties": {
        "test": {
            "type": "number",
            "title": "Test",
            "description": "Testcase",
            "default": 1e-5
        }
    }
}
```

Used commandline:
```
$  datamodel-codegen --input test_codegen.json --output model_test_codegen.py --output-model-type pydantic_v2.BaseModel --input-file-type jsonschema
```

**Observed behavior**
```python
class Model(BaseModel):
    test: Optional[float] = Field('1e-5', description='Testcase', title='Test')

```

**Expected behavior**
```python
class Model(BaseModel):
    test: Optional[float] = Field(1e-5, description='Testcase', title='Test')
```

**Version:**
 - OS: Linux
 - Python version: 3.10.12
 - datamodel-code-generator version: 0.25.6

Maybe related to #1952 .

Issue #1952, opened 2024-05-09
Support scientific notation (YeX)


swelborn:
**Describe the bug**
Scientific notation in json converts to str in model:

```python
class Model(BaseModel):
    hello: str
```

**To Reproduce**

Example json:
```json
{
    "hello": 1e-9
}
```

Used commandline:
```
$ datamodel-codegen --input-file-type json --input gen/test-meta.json --output gen/test_scientific_notation.py
```

**Expected behavior**
Expected float.

**Version:**
 - OS: macOS
 - Python version: 3.11.9
 - datamodel-code-generator version: 0.25.6

PR #1956, opened 2024-05-11
Extend parsing of Scientific Notation to JSON capabilites

maximilian-tech:
Solves Issues https://github.com/koxudaxi/datamodel-code-generator/issues/1955#issue-2290971596 and https://github.com/koxudaxi/datamodel-code-generator/issues/1952#issue-2287818185

It boils down to `yaml` vs. `json` floating point parsing. The `SafeLoader` of `yaml` has to be extended to allow for the same range of  `float` notations like `json`.
[See here for more background](https://stackoverflow.com/questions/30458977/yaml-loads-5e-6-as-string-and-not-a-number).
I copied the SO-solution and it does work (for me)


maximilian-tech:
Hi,

I like your consideration towards older users by not breaking the current behavior.

However, I am not sure how such a flag should be called:
`--parse-json-as-json` or `--extend-yaml-scientifc-notation`?

I really think, that this is a bug, as only now JSON is parsed correctly, and should be the new default behavior.

However, I would be happy either way :)

Thanks for this nice tool!

koxudaxi:
Thank you for agreeing with our opinion.
I prefer `--extend-yaml-scientifc-notation`.
Of course, the cli help should explain it a bit more.

> I really think, that this is a bug, as only now JSON is parsed correctly, and should be the new default behavior.

I agree that this is the case in most cases, but the problem is that being right often gets someone in trouble 😅 
It would be appropriate to change to the default behavior as a breaking change during a major version upgrade. It is pretty legitimate.

maximilian-tech:
> It would be appropriate to change to the default behavior as a breaking change during a major version upgrade.

Agreed. 👍

maximilian-tech:
I just wanted to know what the next steps would be. I really would like to have it in the main branch.

gaborbernat:
Seems stalled. If you pick itUp again and rebase it against main we can reopen.

maximilian-tech:
Reopened in https://github.com/koxudaxi/datamodel-code-generator/pull/2365

PR #1983, opened 2024-06-02
feat: new cli option --use-exact-imports

### LINKED ISSUE #1982: AttributeError: 'FieldInfo' object has no attribute '<EnumName>'
**Describe the bug**
Generating from a schema with an Enum type causes `AttributeError: 'FieldInfo' object has no attribute '<EnumName>'`

**To Reproduce**
File structure after codegen should look like:
```
schemas/
├─ bean.json
├─ bean_type.json
src/
├─ __init__.py
├─ bean.py
├─ bean_type.py
main.py
```

With the schemas defined as follows:
`schemas/bean.json`
```json
{
  "$schema": "https://json-schema.org/draft/2020-12/schema",
  "$id": "bean.json",
  "type": "object",
  "title": "Bean",
  "properties": {
    "beanType": { "$ref": "bean_type.json" },
    "name": { "type": "string" }
  },
  "additionalProperties": false,
  "required": ["beanType", "name"]
}

```

`schemas/bean_type.json`
```json
{
  "$schema": "https://json-schema.org/draft/2020-12/schema",
  "$id": "bean_type.json",
  "title": "BeanType",
  "additionalProperties": false,
  "enum": ["STRING_BEAN", "RUNNER_BEAN", "GREEN_BEAN", "BAKED_BEAN"]
}

```

and

`main.py`
```py
from src.bean import Bean

if __name__ == "__main__":
    pass

```

**Used commandline**
```
$ datamodel-codegen \
    --use-title-as-name \
    --use-standard-collections \ 
    --snake-case-field \
    --target-python-version 3.12 \
    --input schemas \
    --input-file-type jsonschema \
    --output src
```

**Expected behavior**
Exactly what happened, except we should then be able to import and use the generated classes. Instead, an AttributeError is raised.

**Version:**
 - OS: MacOS 14.3.1 (23D60)
 - Python version: 3.12.3
 - Pydantic version: 2.7.2
 - datamodel-code-generator version: 0.25.6

**Additional context**
In the generated file `src/bean.py`, if we manually change `from . import bean_type` to `from .bean_type import BeanType`, and the corresponding usage in the `Bean` class definition, the error disappears. Might be related to #1683 / #1684

alpoi-x:
Adds a new CLI options `--use-exact-imports` (off by default) to import types directly instead of modules. For example:

```py
from . import bean_type

class Bean(BaseModel):
    bean_type: bean_type.BeanType
```

becomes

```py
from .bean_type import BeanType

class Bean(BaseModel):
    bean_type: BeanType
```

Resolves #1982 and potentially #1683


lord-haffi:
It would indeed fix #1683. Also supersedes #1684.

datamodel_code_generator/parser/base.py
```diff
@@ -700,6 +702,11 @@ def __change_from_import(
                     from_, import_ = full_path = relative(
                         model.module_name, data_type.full_name
                     )
+                    if imports.use_exact:
+                        if from_ == '.':
+                            from_ = ''
+                        from_ = f'{from_}.{import_}'
+                        import_ = data_type.reference.short_name
```
lord-haffi: Looks kinda magical though :D

alpoi: True 😅 I'll add some comments and tests

lord-haffi (approved):
lgtm 👍
Feel free to also link #1683 to this PR, I can't do it myself.

datamodel_code_generator/imports.py
```diff
@@ -26,11 +26,12 @@ class Imports(DefaultDict[Optional[str], Set[str]]):
     def __str__(self) -> str:
         return self.dump()
 
-    def __init__(self) -> None:
+    def __init__(self, use_exact: bool = False) -> None:
         super().__init__(set)
         self.alias: DefaultDict[Optional[str], Dict[str, str]] = defaultdict(dict)
         self.counter: Dict[Tuple[Optional[str], str], int] = defaultdict(int)
         self.reference_paths: Dict[str, Import] = {}
+        self.use_exact: bool = use_exact
```
koxudaxi, line 34: Do we need the `self.use_exact` ?

datamodel_code_generator/parser/base.py
```diff
@@ -700,6 +710,10 @@ def __change_from_import(
                     from_, import_ = full_path = relative(
                         model.module_name, data_type.full_name
                     )
+                    if imports.use_exact:
+                        from_, import_ = exact_import(
+                            from_, import_, data_type.reference.short_name
+                        )
```
koxudaxi, line 716: The repo prefer cover almost lines by the unittest.

koxudaxi (commented):
@alpoi-x @lord-haffi 
I left a few comments.
Otherwise, It looks good!!

datamodel_code_generator/imports.py
lord-haffi: It is needed because the method below `__change_from_import` has no access to instance attributes since it is a classmethod. So, the attribute is needed here to propagate the flag all the way down to that classmethod where the flag is actually needed.

datamodel_code_generator/parser/base.py
alpoi: Hiya @koxudaxi, I'm not entirely sure how to test these lines 😓 I can't find any existing tests for this class method (`Parser.__change_from_import`) so I've got nothing to base it on.

lord-haffi: @alpoi-x I think there are tests for executing the whole command. You could just take one of these and test the new option with multi-file-output (I don't remember where this test is but I think, I saw it few months ago). You could then assert that the imports look the way you want.

Issue #1638, opened 2023-10-23
Modular references require output directory error


yogeshVU:
**Describe the bug**
When working with a jsonschema and invoking the model generator, one encounters following error:

`Modular references require an output directory, not a file`

It appears that the naming causes this error          
`        "title": "GenomeStudio 2.0 - Methylation Module",
`

**To Reproduce**

Example schema:
```json

{
  "title": "test",
  "type": "object",
  "properties": {
"Liquid Biomarkers Input Data": {
      "type": "object",
      "properties": {
                "Methylomics": {
          "title": "Methylomics",
          "type": "object",
          "properties": {
            "Software": {
              "title": "Software",
              "type": "object",
              "anyOf": [
                {
                  "title": "GenomeStudio 2.0 - Methylation Module",
                  "type": "object",
                  "properties": {
                    "GenomeStudio": {
                      "title": "GenomeStudio",
                      "type": "object",
                      "properties": {},
                      "required": [],
                      "additionalProperties": false
                    }
                  },
                  "additionalProperties": false
                }

              ]
            }

          },
          "required": [],
          "additionalProperties": false
        }
      },
      "required": []
    }
  },
  "additionalProperties": false,
  "required": []
}
```

Used commandline:
```
$ datamodel-codegen --use-annotated --use-title-as-name --use-subclass-enum --disable-appending-item-suffix --input-file-type jsonschema  --input EnigmaContentModel.json --output testEnigmaModel.py --output-model-type pydantic_v2.BaseModel --allow-population-by-field-name --snake-case-field
```

**Expected behavior**
Should not produce `any modular reference errors`. 

**Version:**
 - OS: [e.g. iOS] macOS
 - Python version: 3.10
 - datamodel-code-generator version: [e.g. 22] 0.22

**Additional context**
Changing the tittle to simple naming eliminates the issue: such as `        "title": "GenomeStudio",
`

koxudaxi:
Thank you for creating the PR.
I guess some users may use the `.` in the title. it's to define a model as a module :thinking: 
Should we provide the option to strip `.` from the title?

Issue #1600, opened 2023-10-05
Incorrect relative imports when using Modular Schema


Timur941:
**Describe the bug**
When generating models from the openapi specification and when the names of the schemas contain dots, incorrect imports are generated inside __init__.py for modules that are actually located in the same directory

Only inside init.py the extra dot in the import statement is added. In other generated .py files, imports are generated correct

**To Reproduce**

Example schema:
```json

{
  "openapi": "3.0.2",
  "info": {
    "title": "Swagger",
    "version": "1.0.17"
  },
  "paths": {
  },
  "components": {
    "schemas": {
      "A.B.C": {
        "type": "object",
        "properties": {
          "id": {
            "type": "integer",
            "format": "int64",
            "example": 10
          },
          "reference": {
            "$ref": "#/components/schemas/A.B.D.E"
          }
        }
      },
      "A.B.D.E": {
        "type": "object",
        "properties": {
          "id": {
            "type": "integer",
            "format": "int64",
            "example": 10
          }
        }
      }
    }
  }
}

```

Used commandline:
```
$ datamodel-codegen --input swagger.json --output model --input-file-type openapi
```
**Actual behavior**
The following file and folder structure was created
```
.
└── model/
    └── A/
        ├── B/
        │   ├── D.py
        │   └── __init__.py
        └── __init__.py
```
Code for init.py inside B folder

```python
# generated by datamodel-codegen:
#   filename:  swagger.json
#   timestamp: 2023-10-05T14:48:13+00:00

from __future__ import annotations

from typing import Optional

from pydantic import BaseModel, Field

from .. import D


class C(BaseModel):
    id: Optional[int] = Field(None, example=10)
    reference: Optional[D.E] = None
```
There are one extra '.' in 'from .. import D' statement because module D is in the same directory

Trying to use a class from that init.py  and getting error
```python
from model.A.B import C

c = C(id=10)

```
```powershell
PS C:\Projects\model-test> python .\test.py
Traceback (most recent call last):
  File "C:\Projects\model-test\test.py", line 1, in <module>
    from model.A.B import C
  File "C:\Projects\model-test\model\A\B\__init__.py", line 11, in <module>
    from .. import D
ImportError: cannot import name 'D' from 'model.A' (C:\Projects\model-test\model\A\__init__.py)
```

**Expected behavior**
Relative Import inside init.py should not contain an extra dot in situations where reference to a module in the same directory is used

For init.py code from above it expected to look like this
```python
[...]

PR #2008, opened 2024-06-18
feature: dots in paths

luca-knaack-webcom:
This PR introduces the `--treat-dot-as-module` flag, allowing users to specify whether a filename like `file.name.of.schema.json` should generate a model at `file_name_of_schema.py` or `file/name/of/schema.py`. This is particularly beneficial for schemas defined as `api.route.path.input.py` and `api.route.path.output.py`, ensuring that the resulting Pydantic models are neatly organized into their respective packages.

Additionally, this PR addresses an issue where filenames containing dots would result in Python model files with the same dots, making them non-importable and non-compliant with PEP naming guidelines.

Merging this PR is likely to resolve issues such as #1638 and #1600.

The PR maintains backward compatibility by introducing new features that are accessible through an optional flag. It only modifies the interpretation of filenames with additional dots and does not affect folders containing dots (e.g., `test_external_relative_ref`).


luca-knaack-webcom:
@koxudaxi, is there anything I can add to this PR to get it merged and resolve the issues?

koxudaxi:
@luca-knaack-webcom 
Thank you for the excellent PR.

This may not be your problem, but it seems that covarage is not meeting 100%. Could you please confirm this? 🤔 
https://github.com/koxudaxi/datamodel-code-generator/pull/2008/checks?check_run_id=26407631809
https://github.com/koxudaxi/datamodel-code-generator/pull/2008/checks?check_run_id=26407425938

The PR should not be in mergeable status if it does not meet 100%. :(

luca-knaack-webcom:
I see. Thanks for pointing that out :)

While adding tests for the missing lines I encountered strange errors in the python-version 3.12 matrix branches of the test.yml workflow. 
Errors were mostly regarding `TypeError: ForwardRef._evaluate() missing 1 required keyword-only argument: 'recursive_guard'` or complaints that pytest wasn't found in the Unit Test step. 
For example: [error-type 1](https://github.com/koxudaxi/datamodel-code-generator/actions/runs/9661391380/job/26649024884?pr=2008) and [error-type 2](https://github.com/koxudaxi/datamodel-code-generator/actions/runs/9661391380/job/26649024662?pr=2008)
Those test-errors did not seem to be related to my code, however, as they still occur now, although I reverted the changes made since last all tests ran.

Did you encounter this before @koxudaxi ?

PR #2009, opened 2024-06-18
Fix Missing Imports

### LINKED ISSUE #2001: Missing Imports When Model Is Changed as a Discriminator
**Describe the bug**
int `Parser.parse` function the `self.__apply_discriminator_type(model, imports)` function gets called `for module, models in module_models`, before `module, models, init, imports, scoped_model_resolver` get added to the `processed_models` array. If certain condition match, this function call has side effects on the already processed models, changing type annotations in the model but not adjusting the imports, thus leading to incorrect models.

**To Reproduce**

Example schemas 
`schema.json`:
```json
{
  "properties": {
    "inner": {
      "discriminator": {
        "mapping": {
          "a": "./type_1.json",
          "A": "./type_1.json"
        },
        "propertyName": "type_"
      },
      "oneOf": [
        {
          "$ref": "./type_1.json"
        }
      ],
      "title": "Inner"
    }
  },
  "required": [
    "inner"
  ],
  "title": "Response",
  "type": "object"
}
```

`type_1.json`:
```json
{
  "properties": {
    "type_": {
      "default": "a",
      "enum": ["a", "A"],

      "type": "string",
      "title": "Type"
    }
  },
  "title": "Type1",
  "type": "object"
}
```

Used commandline:
```
$ datamodel-codegen --input folder/where/the/files/are --output /output --output-model-type pydantic_v2.BaseModel
```

**Expected behavior**
I would expect the resulting pydantic files to import everything they use, but for `type_1.json` I get:
```python

from enum import Enum
from typing import Optional

class Type(Enum):
    a = 'a'
    A = 'A'


class Type1(BaseModel):
    type_: Literal['a', 'A'] = Field(..., title='Type')
```
This model imports `Optional`, which is not used anymore. Optional was deleted [here](https://github.com/koxudaxi/datamodel-code-generator/blob/5727116c36563afbadfe593191286aac50a7b354/datamodel_code_generator/parser/base.py#L820), while handling `schema.json`. Before the `__apply_discriminator_type` method call for `schema.json`, the model for `type_1.json was:
```python
class Type(Enum):
    a = 'a'
    A = 'A'
class Type1(BaseModel):
    type_: Optional[Type] = Field('a', title='Type')
```

and after:
```python
class Type(Enum):
    a = 'a'
    A = 'A'
class Type1(BaseModel):
    type_: Literal['a', 'A'] = Field(..., title='Type')
```
Note, that the `Type1` model of `type_1.json` in the `processed_models` contains the missing import, but also the wrong `Optional` import.

**Version:**
 - OS: macOS
 - Python version: 3.12.3
 - datamodel-code-generator version: using local version with commit [5727116](https://github.com/koxudaxi/datamodel-code-generator/commit/5727116c36563afbadfe593191286aac50a7b354)

**Additional context**
The missing import can be easily added by posprocessing the `processed_models`:
```python
for processed_model in processed_models:
    for model in processed_model.models:
        processed_model.imports.append(model.imports)
```
However, this does not remove the unused model (e.g. `Type` in this example) or unused imports (e.g. `Optional`).

luca-knaack-webcom:
The `processed_models` array stores `Processed` objects, which include models and their imports after being handled by the `parser/base.py:parse` function. However, since `models` are objects (passed by reference), the `Processed` objects and their models can be modified after creation. This can sometimes result in model changes without corresponding updates to the model imports, leading to broken Pydantic models (example is given in #2001).

This PR addresses this issue by introducing a post-processing step in the base parser's parsing method, ensuring that all necessary imports are correctly included.

Fixes #2001


koxudaxi:
Thank you for creating the PR again.
Could you please fix the test coverage?

https://github.com/koxudaxi/datamodel-code-generator/pull/2009/files

luca-knaack-webcom:
@koxudaxi 
Just performed a final code-cleanup. Everything is now covered by tests. While there is one [`pragma: no cover`](https://github.com/koxudaxi/datamodel-code-generator/pull/2009/files#diff-34a68196dba5b4f2cef091ddb835f794178254d4c94bff4a84f84e546f2f6db7R866), this is still covered by [those lines](https://github.com/koxudaxi/datamodel-code-generator/pull/2009/files#diff-34a68196dba5b4f2cef091ddb835f794178254d4c94bff4a84f84e546f2f6db7R1350-R1361) being covered 😅

PR #2014, opened 2024-06-25
patch pydantic v1.10.9 conflict with py3.12

luca-knaack-webcom:
There are errors in pipelines that run the testing configuration `python=3.12` and `pydantic=1.10.9`. The error is something like:
```
| ________________ ERROR collecting tests/parser/test_openapi.py _________________
| tests/parser/test_openapi.py:10: in <module>
|     from datamodel_code_generator import OpenAPIScope, PythonVersion
| datamodel_code_generator/__init__.py:35: in <module>
|     from datamodel_code_generator.parser.base import Parser
| datamodel_code_generator/parser/base.py:37: in <module>
|     from datamodel_code_generator.model import pydantic as pydantic_model
| datamodel_code_generator/model/__init__.py:5: in <module>
|     from ..types import DataTypeManager as DataTypeManagerABC
| datamodel_code_generator/types.py:240: in <module>
|     class DataType(_BaseModel):
| .venv/lib/python3.12/site-packages/pydantic/main.py:286: in __new__
|     cls.__try_update_forward_refs__()
| .venv/lib/python3.12/site-packages/pydantic/main.py:808: in __try_update_forward_refs__
|     update_model_forward_refs(cls, cls.__fields__.values(), cls.__config__.json_encoders, localns, (NameError,))
| .venv/lib/python3.12/site-packages/pydantic/typing.py:554: in update_model_forward_refs
|     update_field_forward_refs(f, globalns=globalns, localns=localns)
| .venv/lib/python3.12/site-packages/pydantic/typing.py:520: in update_field_forward_refs
|     field.type_ = evaluate_forwardref(field.type_, globalns, localns or None)
| .venv/lib/python3.12/site-packages/pydantic/typing.py:66: in evaluate_forwardref
|     return cast(Any, type_)._evaluate(globalns, localns, set())
| E   TypeError: ForwardRef._evaluate() missing 1 required keyword-only argument: 'recursive_guard'
```
and results in a massive amount of errors like:
```
ERROR tests/test_format.py - TypeError: ForwardRef._evaluate() missing 1 required keyword-only argument: 'recursive_guard'
```

To patch the missing `recursive_guard` keyword argument, I created a `pydantic_patch` file that gets imported early in `datamodel_code_generator.__init__.py` and that patches the underlying `evaluate_forwardref` function.

This PR is should fix most of the broken tests in the pipelines for pull requests #2013, #2009 and #2008.

### DIFF
--- a/datamodel_code_generator/pydantic_patch.py
+++ b/datamodel_code_generator/pydantic_patch.py
@@ -1,9 +1,13 @@
+import sys
+
 import pydantic.typing
 
 
 def patched_evaluate_forwardref(forward_ref, globalns, localns=None):
     try:
-        return forward_ref._evaluate(globalns, localns or None, set())  # pragma: no cover
+        return forward_ref._evaluate(
+            globalns, localns or None, set()
+        )  # pragma: no cover
     except TypeError:
         # Fallback for Python 3.12 compatibility
         return forward_ref._evaluate(
@@ -11,4 +15,5 @@ def patched_evaluate_forwardref(forward_ref, globalns, localns=None):
         )
 
 
-pydantic.typing.evaluate_forwardref = patched_evaluate_forwardref
+if '3.12' in sys.version:
+    pydantic.typing.evaluate_forwardref = patched_evaluate_forwardref
--- a/datamodel_code_generator/pydantic_patch.py
+++ b/datamodel_code_generator/pydantic_patch.py
@@ -1,0 +1,14 @@
+import pydantic.typing
+
+
+def patched_evaluate_forwardref(forward_ref, globalns, localns=None):
+    try:
+        return forward_ref._evaluate(globalns, localns or None, set())
+    except TypeError:
+        # Fallback for Python 3.12 compatibility
+        return forward_ref._evaluate(
+            globalns, localns or None, set(), recursive_guard=set()
+        )
+
+
+pydantic.typing.evaluate_forwardref = patched_evaluate_forwardref
--- a/datamodel_code_generator/pydantic_patch.py
+++ b/datamodel_code_generator/pydantic_patch.py
@@ -3,7 +3,7 @@ import pydantic.typing
 
 def patched_evaluate_forwardref(forward_ref, globalns, localns=None):
     try:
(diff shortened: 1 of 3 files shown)


luca-knaack-webcom:
While there is still one failing pipeline, this is the godspeed-benchmark, that isn't able to run on any PR's pipeline.

luca-knaack-webcom:
@koxudaxi, I had to change the `codspeed-benchmark` action to always install the dependencies, as the cached virtual environment seems to lack the `pytest-xdist` package. Once the cache is refreshed, the commented out line could be reintroduced.

datamodel_code_generator/pydantic_patch.py
```diff
@@ -0,0 +1,21 @@
...
+            globalns, localns or None, set()
+        )  # pragma: no cover
+    except TypeError:
+        # Fallback for Python 3.12 compatibility
+        return forward_ref._evaluate(
+            globalns, localns or None, set(), recursive_guard=set()
+        )
+
+
+if '3.12' in sys.version:  # pragma: no cover
```
koxudaxi, line 20: Is this a problem that also occurs in, for example, Python 3.13?

koxudaxi:
@cclauss 
Thank you for spending so much time on this difficult issue.
I was commenting on the code, but I am concerned about the support for other versions, but I don't think it is a problem now, so I will merge it.
Thank you very much.

koxudaxi:
For performance reasons, you may also want to include a version check on the pydantic side.

datamodel_code_generator/pydantic_patch.py
luca-knaack-webcom: I checked this now with python `v3.13.0-beta.2` and pydantic `v2.8.0b1` (earliest version supporting python 3.13) and the problem does not occur. It seems like the broken `ForwardRef` is really only happening for python 3.12

cclauss: https://www.python.org/download/pre-releases/ --> Beta 3 ([Python 3.13.0b3](https://www.python.org/downloads/release/python-3130b3/))

PR #2015, opened 2024-06-26
Fix merging config and args

mahdilamb:
Currently the config and args don't combine if they are sequences or dictionaries. This serves to fix that


koxudaxi:
@mahdilamb 
The unittest failed. Could you please fix it to keep the current behavior?

mahdilamb:
Will do in the next few days! A bit busy at the moment, but it's on my radar!

mahdilamb:
@koxudaxi that should be everything fixed

koxudaxi:
@mahdilamb
I'm sorry for the late reply.
Thank you very much!!

PR #2020, opened 2024-06-28
Fix test coverage

koxudaxi:


koxudaxi:
`  # no: merge` once because pragma is ignored.
https://app.codecov.io/gh/koxudaxi/datamodel-code-generator/pull/2020/blob/datamodel_code_generator/parser/base.py#L1009

PR #2033, opened 2024-07-12
Fix --use-annotated for pydantic v2

bpsoos:
- Fixes two bugs for pydantic v2 --use-annotated:
  - incorrect default value for optional lists
  - incorrect default value "style" for Annotated fields:
     default value style before the fix:
   `fieldname: Annotated[<type>, Field(<default>, <kwargs>)]` 
    the correct style, after the fix:
   `fieldname: Annotated[<type>, Field(<kwargs>)] = <default>`


tests/data/expected/main/jsonschema/null_and_array_v2.py
```diff
@@ -10,7 +10,7 @@
 
 
 class MyObjItem(BaseModel):
-    items: Optional[List[Any]] = None
+    items: Optional[List[Any]]
```
koxudaxi: Will this change alter the behavior?

```python

from typing import Any, List, Optional

class MyObjItem(BaseModel):
    items: Optional[List[Any]]


MyObjItem()
# pydantic_core._pydantic_core.ValidationError: 1 validation error for MyObjItem
# items
#  Field required [type=missing, input_value={}, input_type=dict]
#    For further information visit https://errors.pydantic.dev/2.8/v/missing

```

```python

class MyObjItem(BaseModel):
    items: Optional[List[Any]] = None


MyObjItem()
# no error
```

koxudaxi (changes requested):
@bpsoos
Sorry for the late reply. 
Thank you for the wonderful PR. 
I have commented on one change that concerns me. Please check it.

tests/data/expected/main/jsonschema/null_and_array_v2.py
bpsoos: I think you are right. I have pushed some changes to preserve this behaviour.

However I think this should be fixed in the next breaking release as this is clearly a bug.
`MyObjItem.items` is marked as required in the jsonschema. Also, now we have a discrepancy between the behaviour of the pydantic v1 and v2 models we generate because the v1 one (correctly) doesn't have the default None.

bpsoos:
Hi @koxudaxi,

Thanks for the review! I addressed above the change you were concerned about. Let me know what you think!

koxudaxi (approved):

koxudaxi:
Looks good :)
Thank you very much!!

PR #2043, opened 2024-07-23
Fix missing imports for collapsed models

kmichel-aiven:
When an externally referenced model is collapsed, the required imports for the replacing type(s) were missing.


koxudaxi:
@kmichel-aiven 
I'm sorry for my late response.
Thank you very much!!

PR #2089, opened 2024-09-17
Fix an exact import bug with deeper nesting

AniketDas-Tekky:
If a schema hierarchy was deeper than 1 level and a schema referenced another schema at a higher level, then the exact import function would add an extra '.'.
This change checks if `from_` is any number of '.' and prepends it to the `import_` to maintain correct heirarchy.


docs/development-contributing.md
```diff
@@ -22,7 +22,7 @@ $ cd datamodel-code-generator
 $ curl -sSL https://install.python-poetry.org | python3 - 
 
 ## 3. Install dependencies
-$ poetry install
+$ poetry install --all-extras
```
AniketDas-Tekky, line 25: `tests.sh` fails without graphql installed

AniketDas-Tekky:
@koxudaxi Is there a way to rerun the action? The action failed with "Pytest missing" which seems strange.

prmshepherd:
> @koxudaxi Is there a way to rerun the action? The action failed with "Pytest missing" which seems strange.

I had the same problem in my PR too

datamodel_code_generator/parser/base.py
```diff
@@ -238,11 +238,15 @@ def relative(current_module: str, reference: str) -> Tuple[str, str]:
     return left, right
 
 
+# def absolute()
```
koxudaxi: Can you remove it?

koxudaxi:
> @koxudaxi Is there a way to rerun the action? The action failed with "Pytest missing" which seems strange.

I don't have any idea how to retry it from your side.

OK, I just cleared the cache in GitHub Actions. looks good.

PR #2100, opened 2024-10-09
New CLI option --output-datetime-class #1996

### LINKED ISSUE #1996: Support NaiveDateTime
Add support to choose between `AwareDateTime`, `NaiveDateTime` or generic `datetime`.

A CLI option to choose between the both.

Updating the generated models

**Additional context**
Migrating to pydantic v2 has proven to be very time consuming. Typing becomes more strict, which is a good thing, but going the extra mile as to update the full codebase to ensure all `datetime` objects are using a timezone is too demanding when interact with other tools such as SQLAlchemy, etc.

Pydantic V2 supports:
- `AwareDateTime`
- `NaiveDateTime`
- or the more generic `datetime`

Being able to use the later would make things much easier.

Thanks

archetipo:
to choose between AwareDateTime, NaiveDateTime, generic datetime 
 FIX #1996 and others


meliache:
Thanks for this PR, finally i can move away from the workaround with the Python API and move back to using the CLI directly.

Note that this PR not just introduces an options but also changes the default from `AwareDatetime` to the generic `datetime`.  That's something I would emphasize in the changelog or release notes.

koxudaxi:
@meliache 
Thank you for leaving the comment. I see; I completely overlooked the change in default behavior. If possible, I'd like to maintain the existing behavior. What do you think about that? In other words, I don't want to affect existing users.

Issue #2002, opened 2024-06-12
One `Literal` generated from `discriminator.mapping` doesn't have default


stevapple:
**Describe the bug**
When a `Literal` field is generated from `discriminator.mapping` in an OpenAPI spec (to Pydantic v2), a default value is not generated even with `--use-one-literal-as-default` flag.

**To Reproduce**

Example schema:
```yaml
openapi: "3.0.0"
info:
  version: "0.0.1"
  title: "Demo OpenAPI spec"
paths: {}
components:
  schemas:
    ResponseEvent:
      discriminator:
        propertyName: "type"
        mapping:
          start: "#/components/schemas/StartEvent"
          end: "#/components/schemas/EndEvent"
      oneOf:
        - $ref: "#/components/schemas/StartEvent"
        - $ref: "#/components/schemas/EndEvent"
    StartEvent:
      type: "object"
      properties:
        type:
          type: "string"
      required:
        - type
    EndEvent:
      type: "object"
      properties:
        type:
          type: "string"
      required:
        - type
```

Used commandline:
```
$ datamodel-codegen --input openapi.yml --input-file-type openapi --output model.py --output-model-type pydantic_v2.BaseModel --use-one-literal-as-default
```

Output:
```python
# generated by datamodel-codegen:
#   filename:  openapi.yml
#   timestamp: 2024-06-12T14:20:48+00:00

from __future__ import annotations

from typing import Union

from pydantic import BaseModel, Field, RootModel
from typing_extensions import Literal


class StartEvent(BaseModel):
    type: Literal['start']


class EndEvent(BaseModel):
    type: Literal['end']


class ResponseEvent(RootModel[Union[StartEvent, EndEvent]]):
    root: Union[StartEvent, EndEvent] = Field(..., discriminator='type')
```

**Expected behavior**
The `type` property of `StartEvent` and `EndEvent` should have default values because they are "single literal"s.

**Version:**
 - OS: macOS
 - Python version: 3.12.3
 - datamodel-code-generator version: 0.25.7

**Additional context**

PR #2102, opened 2024-10-11
dataclass generator improvements

### LINKED ISSUE #1966: Dataclass code gen error: TypeError: non-default argument 'XYZ' follows default argument
**Describe the bug**
Generated classes create code that errors with TypeError: non-default argument 'XYZ' follows default argument.

**To Reproduce**
```
Example schema:
openapi: 3.0.0
components:
  schemas:
    Error:
      type: object
      description: Parent error component inherited by all error types.
      properties:
        message:
          type: string
    InternalServerError:
      description: Internal Server Error
      allOf:
        - $ref: '#/components/schemas/Error'
        - type: object
          required: [code]
          properties:
            code:
              $ref: '#/components/schemas/InternalServerErrorCode'
    InternalServerErrorCode:
      type: string
      enum: [INTERNAL_SERVER_ERROR]
    ServiceUnavailableError:
      description: Service Unavailable Error
      allOf:
        - $ref: '#/components/schemas/Error'
        - type: object
          required: [code]
          properties:
            code:
              $ref: '#/components/schemas/ServiceUnavailableErrorCode'
    ServiceUnavailableErrorCode:
      type: string
      enum: [SERVICE_UNAVAILABLE]
```

Used commandline:

```
datamodel-codegen --input .\spec.yml --input-file-type openapi --output .\model.py --output-model-type dataclasses.dataclass --use-generic-container-types --use-generic-container-types --enable-version-header --target-python-version 3.11 --field-constraints --use-annotated  --use-default
```

Generated code:

```

from dataclasses import dataclass
from enum import Enum
from typing import Optional


@dataclass
class Error:
    message: Optional[str] = None


class InternalServerErrorCode(Enum):
    INTERNAL_SERVER_ERROR = 'INTERNAL_SERVER_ERROR'


class ServiceUnavailableErrorCode(Enum):
    SERVICE_UNAVAILABLE = 'SERVICE_UNAVAILABLE'


@dataclass
class InternalServerError(Error):
    code: InternalServerErrorCode


@dataclass
class ServiceUnavailableError(Error):
    code: ServiceUnavailableErrorCode

```
Which is inconsistent and will lead to the following error:

```
TypeError: non-default argument 'code' follows default argument
```

This can be solved by adding kw_only argument as to the dataclass annotation as follows: @dataclass(kw_only=True) as described in:
 https://stackoverflow.com/questions/69711886/python-dataclasses-inheritance-and-default-values

anis-campos:
**Use apply_discriminator_type for dataclasses**
Allow dataclass models to be properly generated with discriminator field

**Fix dataclass inheritance**
Thanks to keyword only, dataclass models can use inheritance and no have issues with default values

**Support datetime types in dataclass fields**

applying `--output-datetime-class` from https://github.com/koxudaxi/datamodel-code-generator/pull/2100 to dataclass to map date, time and date time to the python `datetime` objects instead of strings.

fixes #1966


koxudaxi:
@anis-campos 
Thank you for submitting this pull request. I have two points that need to be addressed:

Since this change alters both the generated code and its behavior, could you please implement the kw_only feature as a new CLI option?
The kw_only parameter is a feature introduced in Python 3.10. We need to ensure that the output version is compatible with this feature. Please refer to the official documentation for more details:
https://docs.python.org/3/library/dataclasses.html#dataclasses.KW_ONLY

I would appreciate it if you could address these points. Thank you for your contribution, and I look forward to reviewing the updated version.

koxudaxi (changes requested):
I left the comment.

anis-campos:
@koxudaxi  I added a cli flag, let me know if the naming is right and if I could have changed less signatures with a different strategy to pass the info down to the model template.

anis-campos:
@koxudaxi  I'm also adding the support for `datetime` in dataclass, I'm using the recently merged `--output-datetime-class` to do so

meliache:
I think this might fix also #2002 where `--use-one-literal-as-default` didn't work for discriminators. At least in my organizations codebase I thought that #2002 is the problem and this PR-branch seems to fix it. Thanks a lot, looking forward very much to see this merged and eventually in a release!

anis-campos:
> I think this might fix also #2002 where `--use-one-literal-as-default` didn't work for discriminators. At least in my organizations codebase I thought that #2002 is the problem and this PR-branch seems to fix it. Thanks a lot, looking forward very much to see this merged and eventually in a release!

I don't think it is linked, the issues is referencing pyantic output, I'm reworking only the dataclass generator.

side note, I would close the issue you linked, executing the example I see no issue with the pydantic output, the I have the default values just fine
```python
#   filename:  openapi.yaml
#   timestamp: 2024-10-15T09:02:19+00:00

from typing import Literal, Union

from pydantic import BaseModel, Field, RootModel


class StartEvent(BaseModel):
    type: Literal['start'] = 'start'


class EndEvent(BaseModel):
    type: Literal['end'] = 'end'


class ResponseEvent(RootModel[Union[StartEvent, EndEvent]]):
    root: Union[StartEvent, EndEvent] = Field(..., discriminator='type')
```

meliache:
@anis-campos
> I don't think it is linked, the issues is referencing pyantic output, I'm reworking only the dataclass generator.

You are right, the linked issue seems to be already resolved in v0.26.1. In my project I had been using v0.25.9 and I had seen the same issue. In the release notes to v0.26.0 and v0.26.1 I saw nothing that indicated that explicitly indicated that the issue might be fixed. In this PR you at least mentioned that you worked on discriminator support, so on a whim I decided to try this branch, but then I tried v0.26.1 and there it also worked already.

Issue #1832, opened 2024-02-02
RuntimeError: Discriminator type is not found


torazem:
**Describe the bug**
When I try to generate models from an OpenAPI schema containing discriminators, I get an error: `RuntimeError: Discriminator type is not found`.

**To Reproduce**
Get the OpenAPI spec for the [Jira REST API](https://developer.atlassian.com/cloud/jira/platform/rest/v3/intro/#version), or the reduced example below. Try to convert it into Pydantic models using `datamodel-code-generator`.

Example schema:
```json
{
    "components": {
        "schemas": {
            "CustomContextVariable": {
                "additionalProperties": false,
                "discriminator": {
                    "mapping": {
                        "issue": "#/components/schemas/IssueContextVariable",
                        "json": "#/components/schemas/JsonContextVariable",
                        "user": "#/components/schemas/UserContextVariable"
                    },
                    "propertyName": "type"
                },
                "oneOf": [
                    {
                        "$ref": "#/components/schemas/UserContextVariable"
                    },
                    {
                        "$ref": "#/components/schemas/IssueContextVariable"
                    },
                    {
                        "$ref": "#/components/schemas/JsonContextVariable"
                    }
                ],
                "properties": {
                    "type": {
                        "description": "Type of custom context variable.",
                        "type": "string"
                    }
                },
                "required": [
                    "type"
                ],
                "type": "object"
            },
            "JsonContextVariable": {
                "description": "A JSON object with custom content.",
                "properties": {
                    "type": {
                        "description": "Type of custom context variable.",
                        "type": "string"
                    },
                    "value": {
                        "description": "A JSON object containing custom content.",
                        "type": "object"
                    }
                },
                "required": [
                    "type"
                ],
                "type": "object"
            },
            "IssueContextVariable": {
                "description": "An [issue](https://developer.atlassian.com/cloud/jira/platform/jira-expressions-type-reference#issue) specified by ID or key. All the fields of the issue object are available in the Jira expression.",
                "properties": {
                    "id": {
                        "description": "The issue ID.",
                        "format": "int64",
                        "type": "integer"
                    },
                    "key": {
                        "description": "The issue key.",
                        "type": "string"
                    },
                    "type": {
                        "description": "Type of custom context variable.",
                        "type": "string"
                    }
                },
                "required": [
                    "type"
                ],
                "type": "object"
            },
            "UserContextVariable": {
[...]

PR #2145, opened 2024-11-05
Fix discriminator mapping resolution in schemas with parent-child hierarchy

sternakt:
This PR addresses [#1832](https://github.com/koxudaxi/datamodel-code-generator/issues/1832), where using discriminators in schemas with specific property mappings can cause a `RuntimeError: Discriminator type is not found`.

### Background

The error occurs because the current discriminator mapping resolution fails to properly handle cases where the discriminator field appears in parent models, preventing the correct mapping key from being located.

### Problem Description

Previously, the code only checked the `discriminator_model` path against the `mapping` keys, but it did not account for scenarios where:

The discriminator path does not directly match the path of the child models, although it matches their parent classes.
Paths need to be validated in a more recursive fashion, checking both parent and child models for a valid mapping.

### Solution

The proposed solution introduces a helper function, `check_paths()`, to validate the paths for both the main `discriminator_model` and its base classes when the initial mapping fails. This function:

- Verifies mapping keys in both the main model and its parent classes.

### Summary

- Introduced `check_paths()` function to handle discriminator mapping resolution.
- Ensured that the discriminator field is validated against both the main model and its parent models, preventing `RuntimeError`.


koxudaxi:
Thank you very much!!

PR #2183, opened 2024-11-22
Add --no-alias

nickyoung-github:
Add a new argument: --no-alias

This is intended  for when you specify a custom base class which has an alias generator but you still want to use --snake-case-field.


nickyoung-github:
Looks like many other outstanding PRs are failing on the pre-commit check. Resource availability?

Issue #2195, opened 2024-11-27
Unused, but requied `target_datetime_class` argument in `get_data_model_types()`


thorwhalen:
**Describe the bug**

It's not a bug, but a code smell that led to errors on my side. 

Simply put: `target_datetime_class` is not used in the [get_data_model_types](https://github.com/koxudaxi/datamodel-code-generator/blob/230fe9b772f1a8018b31b78361f72acf474566ec/datamodel_code_generator/model/__init__.py#L24C5-L24C26) function, and yet it's a required argument.

**To Reproduce**

```python
from datamodel_code_generator import DataModelType, PythonVersion, DatetimeClassType
from datamodel_code_generator.model import get_data_model_types

result = get_data_model_types(
    data_model_type=DataModelType.PydanticBaseModel, 
    target_python_version=PythonVersion.PY_39, 
)
# TypeError                                 Traceback (most recent call last)
# Cell In[15], line 5
#       2 from datamodel_code_generator.model import get_data_model_types
#       4 # Call the function
# ----> 5 result = get_data_model_types(
#       6     data_model_type=DataModelType.PydanticBaseModel, 
#       7     target_python_version=PythonVersion.PY_39, 
#       8 )
# TypeError: get_data_model_types() missing 1 required positional argument: 'target_datetime_class'"
# }
```

**Expected behavior**

I'd expect it to not fail, since the third, `target_datetime_class`, is not used anyway. 

I'd suggest setting a default:
* `target_datetime_class: Optional[DatetimeClassType] = None`, since it's not used anyway. [See this fork branch](https://github.com/thorwhalen/datamodel-code-generator/blob/12da21d9dd4bb311b59184becbd40f4d7f449d0f/datamodel_code_generator/model/__init__.py#L24)
* `target_datetime_class: DatetimeClassType = DatetimeClassType.Datetime`, if we want an actual useful choice. [See this fork branch](https://github.com/thorwhalen/datamodel-code-generator/blob/9ed9a073c590853dd8b39bb93e0283918aeb0e8b/datamodel_code_generator/model/__init__.py#L24)

In fact -- personally, I'd set a default for the `target_python_version` as well, setting it to the python version where the module is being run. 
Yes, explicit over implicit, and having defaults has disadvantages, always. It does a lot for UX though.
Personally, I'd set these defaults in the `format.py` module, where the `DataModelType` and `PythonVersion` enums are defined, and use them throughout. 

[...]

PR #2196, opened 2024-11-27
Default datetime class and python version

thorwhalen:
Solution for issue #2195.


koxudaxi:
Thank you very much!!

PR #2214, opened 2024-12-10
Fix OpenAPI test case failure in fresh clone

ncoghlan:
Following the initial setup guide resulted in one of the test cases failing with this difference in the produced output.


koxudaxi:
Thank you very much!!

PR #2215, opened 2024-12-10
Avoid sharing extra template state between models

### LINKED ISSUE #2211: Extra template data conflicts with `msgspec` discriminated union support
I am attempting to use the `msgspec` support for the OpenAPI `discriminator` fields in JSON Schema (as added in https://github.com/koxudaxi/datamodel-code-generator/commit/da7d0542b333cdb473bf1d7c4bde608401460f9b)

I've been reverse engineering the expected usage from the test suite, but have hit a point where the behaviour is *so* strange it feels like a bug:

* even though only a handful of union fields have `discriminator` set, *every* Struct subclass in the generated file has `tag_field="type:` and `tag=None` (even the ones that have nothing to do with the updated union fields)
* the classes that are legitimately part of the discriminated unions also have `tag=None` set, and their `type` field is just `Literal["value"]` rather than the expected class-var-with-default-value that the test suite says I should be getting

I did notice the test suite used `type_` instead of `type` as its discriminator, so I attempted to set `{"type": "type_"}` as the alias map, but that didn't make any difference to the output. I can't change the field name in the schema itself, as that's defined by the API I'm writing a client for, not by my client code.

I also couldn't find anything in the main documentation indicating that `type` was problematic as a field name.

The `generate` call looks like this (`kw_only` is set via the extra template data because setting the `keyword_only=True` call option didn't have any effect):

```python
    generate(
        _SCHEMA_PATH,
        input_file_type=InputFileType.JsonSchema,
        output=_MODEL_PATH,
        output_model_type=DataModelType.MsgspecStruct,
        field_constraints=True,
        use_annotated=True,
        use_double_quotes=True,
        use_subclass_enum=True,
        use_union_operator=True,
        extra_template_data=defaultdict(
            dict,
            {
                "#all#": {
                    "base_class_kwargs": {
                        # "None" fields should be omitted, not sent as "null"
                        "omit_defaults": True,
                        # Allow non-default fields after default fields
                        "kw_only": True,
                    }
                }
            },
        ),
    )
```

And a snippet of the `"definitions"` entry in the input schema:

```
    "chatMessagePartData": {
      "anyOf": [
        {
          "$ref": "#/definitions/chatMessagePartTextData"
        },
        {
          "$ref": "#/definitions/chatMessagePartFileData"
        },
        {
          "$ref": "#/definitions/chatMessagePartToolCallRequestData"
        },
        {
          "$ref": "#/definitions/chatMessagePartToolCallResultData"
        }
      ],
      "discriminator": {
        "mapping": {
          "text": "#/definitions/chatMessagePartTextData",
          "file": "#/definitions/chatMessagePartFileData",
          "toolCallRequest": "#/definitions/chatMessagePartToolCallRequestData",
          "toolCallResult": "#/definitions/chatMessagePartToolCallResultData"
        },
        "propertyName": "type"
      }
    },
    "chatMessagePartFileData": {
      "type": "object",
      "properties": {
        "type": {
          "type": "string",
          "const": "file"
        },
        "name": {
          "type": "string"
        },
        "identifier": {
          "type": "string"
        },
        "sizeBytes": {
          "type": "number"
        },
        "fileType": {
          "$ref": "#/definitions/fileType"
        }
      },
      "required": [
        "type",
        "name",
        "identifier",
        "sizeBytes",
        "fileType"
      ],
      "additionalProperties": false
    },
```

ncoghlan:
Closes #2211


koxudaxi:
Thank you very much!!

ncoghlan:
And thanks to you, too!

Even with the time needed to resolve this, using `datamodel-code-generator` has been orders of magnitude faster than trying to write my own data model class generator from scratch.

(and thanks to @indrat as well - I've been relying heavily on their recent enhancements to the msgspec support).

PR #2240, opened 2024-12-23
typo: Fix 'Allow to' sentence.

jas4711:
Hi!  I'm packaging this project for inclusion into Debian, and our QA tools found a small typo nit:

```
I: datamodel-codegen: typo-in-manual-page "Allow to" "Allow one to" [usr/share/man/man1/datamodel-codegen.1.gz:155]
```

How about this patch?

Debian packages is available here: https://salsa.debian.org/python-team/packages/python-datamodel-code-generator/

Thanks,
/Simon

### DIFF
--- a/README.md
+++ b/README.md
@@ -435,7 +435,7 @@ Field customization:
                         Use schema description to populate field docstring
 
 Model customization:
-  --allow-extra-fields  Allow to pass extra fields, if this flag is not passed, extra fields
+  --allow-extra-fields  Allow passing extra fields, if this flag is not passed, extra fields
                         are forbidden.
   --allow-population-by-field-name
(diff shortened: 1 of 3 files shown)


koxudaxi:
Thank you for the PR.
I have been using Debian for a very long time. Does this mean it will be available to install via apt-get install?

jas4711:
Thank you @koxudaxi 

Yes, I hope that the Debian ftp-masters will accept this package, and then you should be able to `apt-get install datamodel-codegen` and the tool should be working.  Your package is in great shape, so I don't have any more ideas or suggestions as a result of the packaging work.  The package review process may take weeks or even months, so don't hold your horses.

/Simon

PR #2271, opened 2025-01-16
Bump default target Python version to 3.9

ncoghlan:
Python 3.8 reached end-of-life in October 2024.

`typing.Annotated` was added in Python 3.9, so using annotations will no longer need `typing_extensions` by default.

### DIFF
--- a/datamodel_code_generator/__init__.py
+++ b/datamodel_code_generator/__init__.py
@@ -237,7 +237,7 @@ def generate(
     input_file_type: InputFileType = InputFileType.Auto,
     output: Optional[Path] = None,
     output_model_type: DataModelType = DataModelType.PydanticBaseModel,
-    target_python_version: PythonVersion = PythonVersion.PY_38,
+    target_python_version: PythonVersion = PythonVersion.PY_39,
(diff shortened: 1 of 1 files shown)


ncoghlan:
Not an urgent change, I just got bitten by other projects no longer depending on `typing-extensions` in Python 3.13, which meant I noticed my own dependency that I had inadvertently picked up by generating code compatible with a Python version I didn't support anyway.

ncoghlan:
Reverted to draft, as the CI failure indicates there's more to this change than just bumping the default value.

gaborbernat:
Yeah, this will trigger a lot of changes 😊

koxudaxi:
I always struggle with deciding when to change the drop and default versions 😅

gaborbernat:
> I always struggle with deciding when to change the drop and default versions 😅

My rule of thumb:

- drop when becomes EOL.
- adopt when there's <1 month until new version release (so rc releases available, and all dependencies work).

gaborbernat:
Seems stalled, will close, we'll address this separately.

Issue #2220, opened 2024-12-14
Migrate from Poetry to uv package manager


koxudaxi:
**Description**

We should consider migrating our dependency management from Poetry to uv, which is a new extremely fast Python package installer and resolver written in Rust.

**Benefits of migration:**
- Significantly faster package installation (up to 10-100x faster than pip)
- Built-in compile cache for faster repeated installations
- Reliable dependency resolution
- Native support for all standard Python package formats
- Compatible with existing `pyproject.toml` configurations
- Lower memory usage compared to Poetry

**Required Changes:**
1. Remove Poetry-specific configurations while keeping the essential `pyproject.toml` metadata
2. Update CI/CD pipelines to use uv instead of Poetry
3. Update development setup instructions in documentation
4. Ensure all development scripts/tools are compatible with uv
5. Update the project's virtual environment handling

**Notes for Implementation:**
- uv can read dependencies directly from `pyproject.toml`
- The migration should be backward compatible until fully tested
- We should provide clear migration instructions for contributors
- Consider adding both Poetry and uv support during a transition period

**Resources:**
- uv Documentation: https://docs.astral.sh/uv/
- Migration Guide: https://docs.astral.sh/uv/migration/

**Testing Requirements:**
- Verify all dependencies are correctly resolved
- Ensure development workflow remains smooth
- Test CI/CD pipeline with new setup
- Verify package installation in clean environments

bhargavyagnik:
@koxudaxi , can i take this up?

gaborbernat:
Opened a PR resolving this.

gaborbernat:
Considering this is an open source project and not a paid company project the priority always is what people are willing to work on rather than what problem you are experiencing. As someone who has tried using poetry behind a corporate firewall I can tell you that you that uv works much better, and increases the chances that I am willing to do further contributions.

PR #2288, opened 2025-01-29
Migrate to use uv instead of poetry

gaborbernat:
Use hatchling instead of poetry for backend, uv otherwise.

Resolves https://github.com/koxudaxi/datamodel-code-generator/issues/2220


gaborbernat:
@koxudaxi any update on this? I have another bug I'd like to addrtess, and this is blocking me 😊

gaborbernat:
Ah, I slighthly renamed the GitHub actions so shos more info in my browser without needing to click into it, which causes the missing checks. Should I restore the old naming or you OK updating the required checks?

gaborbernat:
Hmm, unsure why the coverage decreased, I haven't touched any of those files 🤔

gaborbernat:
Orthogonal to this change set, but given that `uv` does not yet support a task manager, would you be open to add `tox-uv` to this project, so easier to setup these envs locally? 😊 this would replace the shell scripts also. `uv` plans to add task manager support later this year at which point we could swap over.

gaborbernat:
> Regarding tox-uv, while it's welcome, I will also read the documentation tomorrow myself.

Disclaimer I am the auther and maintainer of both `tox` and `tox-uv`; so if you have any questions you can ask 😊 Yeah I saw the same, so will need to check. Will put in a commit to add `tox-uv` (as helps me to try to reproduce this easier locally) and circle back.

koxudaxi:
The coverage was fixed after re-running only that specific task.
It's a very silly phenomenon😣
Thank you. I'll be back tomorrow to handle this and other PRs.
I'll leave another message here or somewhere else.
Thank you very much for your cooperation!!

gaborbernat:
Another orthogonal question, would you be open to switch to a `src` layout? 😆 https://packaging.python.org/en/latest/discussions/src-layout-vs-flat-layout/

skshetry:
This PR seems to have changed a lot of dependencies that were optional before:
```toml
prance = { version = ">=0.18.2", optional = true }
openapi-spec-validator = { version = ">=0.2.8,<0.7.0", optional = true }
PySnooper = { version = ">=0.4.1,<2.0.0", optional = true }
httpx = { version = "*", optional = true }
graphql-core = {version = "^3.2.3", optional = true}
```

Is this intentional or an oversight?

skshetry:
@gaborbernat, PTAL at [Extras - poetry](https://python-poetry.org/docs/pyproject#extras).

My understanding reading that is, in poetry, you use it to specify version constraints for optional dependency, which later you are going to use in `extras`. And looks like it's a deprecated way of specifying extra/optional dependencies.

I think you could simply get rid of those dependencies from `project.dependencies` as they are already set in `project.optional-dependencies`.

koxudaxi:
@skshetry 
Thank you for your report. 
I'm working on the PR

https://github.com/koxudaxi/datamodel-code-generator/pull/2306

koxudaxi:
OK, I have released the fixed version `0.27.2`

PR #2289, opened 2025-01-30
Support isort 6

cjwatson:
The breaking changes in isort 6 are just to remove support for Python 3.8 (see https://github.com/PyCQA/isort/releases), so there's no need for this project to treat it differently from isort 5.


cjwatson:
I see the pre-commit failure, but it seems to be entirely due to running with the fact that the check now runs with poetry-core 2.0.1, rather than anything to do with my changes.  Do you want me to include the results of running `poetry lock` anyway?

cjwatson:
Hm, since pre-commit passed in #2290, I guess I have no choice.  I've updated `poetry.lock`.

gaborbernat (changes requested):
Please rebase, to use `uv`.

gaborbernat:
@koxudaxi can you give me admin access to the repo, I'd like to change some settings (enable auto-merge, auto-delete  branches, and such 🤔 ) - you can take it away after I make the changes 😆

gaborbernat (approved):

koxudaxi:
I can't find the option in settings 🤔 
image

gaborbernat:
Ah, because this is under your personal profile and not an organization 😆

koxudaxi:
Oh, OK. I found the document.
https://docs.github.com/en/account-and-profile/setting-up-and-managing-your-personal-account-on-github/managing-user-account-settings/permission-levels-for-a-personal-account-repository

PR #2290, opened 2025-01-30
Include tests in sdist

cjwatson:
Including tests would make it easier for downstream distributors (e.g. Debian) to ensure that changes to other dependencies haven't broken this package.


gaborbernat:
We moved over to hatchling that should have fixed this.

PR #2292, opened 2025-01-31
Add tox configuration file

gaborbernat:
Signed-off-by: Bernát Gábor <bgabor8@bloomberg.net>

PR #2295, opened 2025-02-01
Migrate mypy to pyright

koxudaxi:
Related Issues: https://github.com/koxudaxi/datamodel-code-generator/pull/2292

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -95,6 +95,8 @@ jobs:
       - uses: actions/checkout@v4
         with:
           fetch-depth: 0
+      - name: Install the latest version of uv
+        uses: astral-sh/setup-uv@v5
       - name: Setup check suite
         run: tox r -vv --notest --skip-missing-interpreters false -e ${{ matrix.tox_env }}
       - name: Run check for ${{ matrix.tox_env }}
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -74,3 +74,40 @@ jobs:
         env_vars: OS,PY,ISORT
       env:
         CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
+  check:
+    name: tox env ${{ matrix.tox_env }} on ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        tox_env:
+          - type
+          - dev
+          - docs
+          - pkg_meta
+          - readme
+        os:
+          - ubuntu-latest
+          - windows-latest
+        exclude:
+          - { os: windows-latest, tox_env: docs }
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - name: Install the latest version of uv
+        uses: astral-sh/setup-uv@v5
+        with:
+          enable-cache: true
+          cache-dependency-glob: "pyproject.toml"
+          github-token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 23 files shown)


pyproject.toml
```diff
@@ -77,6 +77,7 @@ test = [
 ]
 type = [
   "mypy>=1.4.1,<1.5",
```
gaborbernat: Shall we then drop mypy from here entirely, and update here https://github.com/koxudaxi/datamodel-code-generator/blob/main/tox.ini#L64 to invoke pyright instead. And we should add a new environment to run it as part of the CI. I tend to add them as another batch of jobs to run alongside tests, see example https://github.com/tox-dev/tox/blob/main/.github/workflows/check.yaml#L58-L93 

We can then add the pkg-meta, readme and type environments there. That way all checks are done part of the CI run.

.github/workflows/test.yml
```diff
@@ -74,3 +74,40 @@ jobs:
...
+          - windows-latest
+        exclude:
+          - { os: windows-latest, tox_env: docs }
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - name: Install the latest version of uv
+        uses: astral-sh/setup-uv@v5
+        with:
```
gaborbernat: Because we are using a lock file and by default the cache is enabled, you can remove the following three lines.

.github/workflows/test.yml
```diff
@@ -74,3 +74,40 @@ jobs:
...
+          enable-cache: true
+          cache-dependency-glob: "pyproject.toml"
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+      - name: Add .local/bin to Windows PATH
```
gaborbernat: This has been fixed in upstream and can be also removed.

.github/workflows/test.yml
```diff
@@ -74,3 +74,40 @@ jobs:
...
+        if: runner.os == 'Windows'
+        shell: bash
+        run: echo "$USERPROFILE/.local/bin" >> $GITHUB_PATH
+      - name: Install tox
```
gaborbernat, line 101: This also can be removed. It's only added there because we're trying to eat our own dog food. But this is managed by UV now.

gaborbernat (changes requested):

gaborbernat (approved):

.github/workflows/test.yml
```diff
@@ -74,3 +74,33 @@ jobs:
...
+      matrix:
+        tox_env:
+          - type
+          - dev
+          - docs
+          - pkg_meta
+          - readme
+        os:
+          - ubuntu-latest
+          - windows-latest
```
gaborbernat, line 91: Do we need Windows at all here? Other than for the `type` environment?

datamodel_code_generator/__main__.py
```diff
@@ -427,7 +429,7 @@ def main(args: Optional[Sequence[str]] = None) -> Exit:
         print(
             f"Installed black doesn't support Python version {config.target_python_version.value}.\n"  # type: ignore
             f'You have to install a newer black.\n'
-            f'Installed black version: {black.__version__}',
+            f'Installed black version: {black.__version__}',  # pyright: ignore [reportPrivateImportUsage]
```
gaborbernat, line 432: IMHO we should disable reportPrivateImportUsage, this is already handled by ruff.

koxudaxi:
Thank you for your comment.
I have opened the new PR.
https://github.com/koxudaxi/datamodel-code-generator/pull/2298

PR #2299, opened 2025-02-03
fix: merge test coverage before upload

koxudaxi:


.github/workflows/test.yml
```diff
@@ -63,17 +63,36 @@ jobs:
...
+    needs: test
+    steps:
+      - name: Download all coverage artifacts
+        uses: actions/download-artifact@v4
+      - name: Install the latest version of uv
+        uses: astral-sh/setup-uv@v5
+      - name: Combine coverage reports and generate XML
+        run: |
+          ls -la
+          uv run coverage combine
```
gaborbernat: You need to checkout the repository, afaik coverage uses the source tree to do the merge.

gaborbernat: You need to tell coverage what files to combine 🤔 setting the env var `COVERAGE_FILE = coverage` I think will do, because then does prefix matching. Probably there's some CLI flag to. I'd instead recommend creating a tox task for it, similar to how https://github.com/tox-dev/filelock/blob/main/tox.ini#L56-L82 does it, to also allow doing this locally.

koxudaxi: Thank you. I'm porting the github action here.

gaborbernat: Yeah, was just proposing to port it to tox, and then just invoke that in GitHub action; that way is re-usabel locally 🤔 up to you though.

tox.ini
```diff
@@ -58,6 +58,33 @@ commands =
     check-wheel-contents --no-config {env_tmp_dir}
 dependency_groups = pkg-meta
 
+[testenv:coverage]
+description = combine coverage files and generate diff (against DIFF_AGAINST defaulting to origin/main)
+skip_install = true
+deps =
+    covdefaults>=2.3
+    coverage[toml]>=7.6.10
+    diff-cover>=9.2.1
```
gaborbernat: Would recommend moving it into a dependency group and depending on it as such: 
https://github.com/koxudaxi/datamodel-code-generator/blob/main/pyproject.toml#L66

.github/workflows/test.yml
koxudaxi: I agree. Adding task makes sense since I've wanted to check integrated coverage locally.
Quick question - is it possible to create a matrix in tox not just for Python versions but also for different versions of pydantic and isort? Some of our code depends on specific library versions, so it would be helpful to run all tests in parallel locally rather than just in CI.
Plus, this would allow us to check coverage as well.

tox.ini
koxudaxi: Fixed!!

pyproject.toml
```diff
@@ -89,6 +90,11 @@ docs = [
 ]
 fix = [ "pre-commit>=3.5" ]
 pkg-meta = [ "check-wheel-contents>=0.6.1", "twine>=6.1", "uv>=0.5.22" ]
+coverage = [
+  "covdefaults>=2.3",
```
gaborbernat: You can remove this for now, as we don't use it in this project, though we should add it at some point to the core test suite, because allows you to do conditional coverage markers 😊

.github/workflows/test.yml
gaborbernat: Yeah, possible, just makes the test matrix a lot bigger. Do we want to add only what we run in the CI?

gaborbernat: Kinda orthogonal, but any reason we want to support all these combinations? I can understand maybe pydantic, but for black and isort I would say latest is the only supported version. How many people use these older versions?

pyproject.toml
koxudaxi: I'm interested, but I'll remove it for now!!

gaborbernat (approved):

.github/workflows/test.yml
```diff
@@ -62,18 +62,64 @@ jobs:
...
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - name: Install the latest version of uv
+        uses: astral-sh/setup-uv@v5
+        with:
+          enable-cache: true
+          cache-dependency-glob: "pyproject.toml"
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+      - name: Install hatch
```
gaborbernat, line 92: `Install tox` 👍

gaborbernat (commented):
Missed this 🤦

koxudaxi:
@gaborbernat 
Although we installed a specific version of Pydantic, it appears that tox is reinstalling the version of pydantic specified in the lock file through uv. At least in the coverage report, there's no evidence that Pydantic 1.x was executed.
image

https://github.com/koxudaxi/datamodel-code-generator/actions/runs/13142814968/job/36673794376#step:7:12

gaborbernat:
Ah I see, let me take a look at this. Can you add me to PyPI, or setup trusted publishing in the meantime?

gaborbernat:
Seems this is a feature gap, added https://github.com/tox-dev/tox-uv/pull/168

.github/workflows/test.yml
gaborbernat: > I agree. Adding task makes sense since I've wanted to check integrated coverage locally.
> Quick question - is it possible to create a matrix in tox not just for Python versions but also for different versions of pydantic and isort? Some of our code depends on specific library versions, so it would be helpful to run all tests in parallel locally rather than just in CI.

So this is not possible with a lock file... because, well the two different dependency groups conflict each other. But we could not use the lock file for this case...

koxudaxi:
> Ah I see, let me take a look at this. Can you add me to PyPI, or setup trusted publishing in the meantime?

OK. Could you tell me your user name in PyPI?

koxudaxi:
Or, Do you recommend to use `trusted publishing` settings ?

gaborbernat:
> Or, Do you recommend to use `trusted publishing` settings ?

later 😎

PR #2303, opened 2025-02-05
setup-trusted-publisher

koxudaxi:
I setup the config.
Also, I add the config to PyPI console.
https://docs.pypi.org/trusted-publishers/using-a-publisher/

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -42,7 +42,7 @@ jobs:
           name: ${{ env.dists-artifact-name }}
           path: dist/
       - name: Publish to PyPI
-        uses: pypa/gh-action-pypi-publish@1.12.4
+        uses: pypa/gh-action-pypi-publish@v1.12.4
         with:
           attestations: true
 
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -30,9 +30,7 @@ jobs:
     needs:
       - build
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yml
```diff
@@ -30,9 +30,7 @@ jobs:
     needs:
       - build
     runs-on: ubuntu-latest
-    environment:
-      name: release
-      url: https://pypi.org/project/datamodel-code-generator/${{ github.ref_name }}
```
gaborbernat, line 35: Keep the url

gaborbernat (changes requested):

.github/workflows/publish.yml
koxudaxi: I reverted it. Thank you.

.github/workflows/publish.yml
```diff
@@ -42,15 +42,15 @@ jobs:
           name: ${{ env.dists-artifact-name }}
           path: dist/
       - name: Publish to PyPI
-        uses: pypa/gh-action-pypi-publish@v1.12.4
+        uses: pypa/gh-action-pypi-publish@release/v1
```
gaborbernat: I prefer keeping this pinned to a version and letting dependabot bump it when there's new changes, but I'm ok with as is it too.

gaborbernat (approved):

gaborbernat (approved):

PR #2311, opened 2025-02-07
Use src layout

gaborbernat:
Also adds coverage to the test suite because:

- https://nedbatchelder.com/blog/201106/running_coverage_on_your_tests.html
- https://www.youtube.com/watch?v=70T6OxKwxm0

Drop contexts for coveragepy as made the combine very slow, once we figure it out why this happens we can add it back.

Also adds https://github.com/asottile/covdefaults


gaborbernat:
@koxudaxi can we merge this as is? The coverage decreased but only because we included tests. I would even suggest to remove the codecov checks from being required, the combine CI step already does everything codecov does when it comes to gate keeping.

koxudaxi:
OK, no problem to remove it 😄

koxudaxi (approved):

koxudaxi:
@gaborbernat 
Thank you very much!!

PR #2318, opened 2025-02-11
Add more ruff checks and use defaults

gaborbernat:
Adopt further ruff rules and fix failures.


gaborbernat:
@koxudaxi this is ready now, is a bit chunky PR, and reqired moving the CI for `py3.9-black19` to `py3.8-black19` because of https://github.com/pydantic/pydantic/pull/3403, but otherwise should be fairly no op (for this you'll need to update the CI check).

One (perhaps) contentious change is that changed the default quote from ' to " because that's the default,  and IMHO defaults are better 👍 let me know if you disagree.

pyproject.toml
```diff
@@ -83,7 +83,7 @@ test = [
   "pytest-cov>=5",
   "pytest-mock>=3.14",
   "pytest-xdist>=3.3.1",
-  "setuptools; python_version<'3.9'",
+  "setuptools; python_version<'3.10'", # PyCharm debugger needs it
```
koxudaxi, line 86: 👀

gaborbernat: I know! Uses `pkg_resources` from setuptools...

koxudaxi:

>. One (perhaps) contentious change is that changed the default quote from ' to " because that's the default, and IMHO defaults are better 👍 let me know if you disagree.

 I was thinking of making the change to match my other projects, so this works out well. thank you 😄

koxudaxi (approved):

PR #2323, opened 2025-02-13
Graphql --set-default-enum-member

andrew-womeldorf:
Adding `--set-default-enum-member` functionality to the graphql parser


gaborbernat (changes requested):
We should document this somewhere under https://koxudaxi.github.io/datamodel-code-generator/graphql, explain why someone would want.

andrew-womeldorf:
> We should document this somewhere under https://koxudaxi.github.io/datamodel-code-generator/graphql, explain why someone would want.

It's already documented with the other parameters, [here](https://koxudaxi.github.io/datamodel-code-generator/#all-command-options). Are you thinking we should add documentation to describe the behavior?

FWIW - I'm adding this to silence the typechecker.

gaborbernat:
> FWIW - I'm adding this to silence the typechecker.

Can you explain. There's no issue linked to explain the context, so not sure.

andrew-womeldorf:
Yes. With the behavior before this change, the output file looks like this:

```python
#   filename:  field-default-enum.graphql
#   timestamp: 2019-07-26T00:00:00+00:00

from enum import Enum
from typing import List, Literal, Optional, TypeAlias

Boolean: TypeAlias = bool
"""
The `Boolean` scalar type represents `true` or `false`.
"""


String: TypeAlias = str
"""
The `String` scalar type represents textual data, represented as UTF-8 character sequences. The String type is most often used by GraphQL to represent free-form human-readable text.
"""


class Color(Enum):
    BLUE = 'BLUE'
    GREEN = 'GREEN'
    RED = 'RED'


class Car(BaseModel):
    colorList: Optional[List[Color]] = ['RED']
    colorOne: Optional[Color] = 'GREEN'
    typename__: Optional[Literal['Car']] = Field('Car', alias='__typename')
```

Where the default values are `Literal`s. Running a type checker throws errors like the following:

```
$ uv tool run pyright example.py
/Users/awomeldorf/tmp/pyright/example.py
  /Users/awomeldorf/tmp/pyright/example.py:10:6 - error: Import "pydantic" could not be resolved (reportMissingImports)
  /Users/awomeldorf/tmp/pyright/example.py:31:41 - error: Type "list[str]" is not assignable to declared type "List[Color] | None"
    "Literal['RED']" is not assignable to "Color" (reportAssignmentType)
  /Users/awomeldorf/tmp/pyright/example.py:32:33 - error: Type "Literal['GREEN']" is not assignable to declared type "Color | None"
    Type "Literal['GREEN']" is not assignable to type "Color | None"
      "Literal['GREEN']" is not assignable to "Color"
      "Literal['GREEN']" is not assignable to "None" (reportAssignmentType)
3 errors, 0 warnings, 0 informations
```

The change made here makes those Enum members instead of Literal.

gaborbernat:
Sounds to me like we definitely should have a test for it.

gaborbernat (approved):

gaborbernat:
Can you fix the type checker before we can merge this.

andrew-womeldorf:
I'm sorry, I think I'm still not communicating this very clearly.

The default behavior of datamodel-code-generator outputs code which fails the typechecker, in the way that I showed in the example above.

The change made in this MR makes the output correctly typed.

Before this change:

```python
...
class Car(BaseModel):
    colorList: Optional[List[Color]] = ['RED']
    colorOne: Optional[Color] = 'GREEN'
```

After this change

```python
class Car(BaseModel):
    colorList: Optional[List[Color]] = [Color.RED]
    colorOne: Optional[Color] = Color.GREEN
```

andrew-womeldorf:
One of the closest issues I found regarding this is https://github.com/koxudaxi/datamodel-code-generator/issues/2255. I had the same expected behavior and issue as described there. [The PR that closed that issue](https://github.com/koxudaxi/datamodel-code-generator/pull/2259) fixed the problem, but it required for references to be used in the DataType. References are already in use in the jsonschema parser, but were not in use in the graphql parser.

gaborbernat:
Yeah I understand now, but before I can accept this you need to fix that failing CI check :D 
image

src/datamodel_code_generator/parser/graphql.py
```diff
@@ -389,7 +389,7 @@ def parse_field(
             obj = obj.of_type  # pyright: ignore[reportAttributeAccessIssue]
 
         if graphql.is_enum_type(obj):
-            data_type.reference = self.references[obj.name]
```
gaborbernat: The error raised a valid issue in your code. I think you need to handle the error, not ignore it. 🤔 Ignore should be only used as last resort, not the first thing you reach for.

andrew-womeldorf: I was following the pattern of the surrounding code. I will add either a conditional or assert

gaborbernat: You probably should include it in the if statement above the conditional.

andrew-womeldorf: It's actually a little interesting... 

- `graphql.is_enum_type()` is supposed to return a `TypeGuard[GraphQLEnumType]`.
- `GraphQLEnumType` is a subclass of `GraphQLNamedType`
- `GraphQLNamedType` has the required `name` field.

So `graphql.is_enum_type()` should have already resolved this.

I've also tried adding `... and graphql.is_named_type(obj)` to be more specific (also returns a TypeGuard), and I've tried `graphql.assert_named_type()`, and the type checker still complains.

I've added a more direct `assert`, which is actually appeasing the typechecker. Let me know if you'd prefer something different.

andrew-womeldorf: https://github.com/graphql-python/graphql-core/blob/main/src/graphql/type/definition.py#L1190

gaborbernat: Could it be a pyright bug we should submit upstream? Can you reproduce it in a small mock case?

If so submit the issue to https://github.com/microsoft/pyright/issues and link it before the assert, so we know why it's there 🤔

andrew-womeldorf: Good idea. Let me see what I can come up with.

andrew-womeldorf: I am stuck on trying to reproduce this issue with my own code. However, I did get a better idea of what's happening.

- `obj = field.type`
- `field.type: graphql.GraphQLInputType`
- `graphql.GraphQLInputType = Union[GraphQLScalarType, GraphQLEnumType, GraphQLInputObjectType, GraphQLWrappingType]`

`graphql.is_enum_type(obj)` does an `isinstance(obj, graphql.GraphQLEnumType)`. However, `obj` can still also be `graphql.WrappingType` (generic) while also being `graphql.GraphQLEnumType`.

`graphql.type.definitions` has several `assert_X_type()` functions, which cast the type to `X`. If I had to guess, those functions exist to solve for the issue faced here.

Therefore, I've replaced my assert statement with `assert_enum_type()` to cast to the known type. I've done similarly for the other two type ignore comments that were directly above and below the code I've been working on, to cast to the known types.

andrew-womeldorf: In case you're curious, here's the work I've got in trying to reproduce the error.

Reproduce the error with graphql code:

```python
import graphql

def parse_field(
    field: graphql.GraphQLInputField,
) -> None:
    if graphql.is_enum_type(field.type):
        print(field.type.name)
```

Fix the error with `graphql.assert_enum_type()`:

```python
import graphql

def parse_field(
    field: graphql.GraphQLInputField,
) -> None:
    if graphql.is_enum_type(field.type):
        field.type = graphql.assert_enum_type(field.type)
        print(field.type.name)
```

My attempt to reproduce the error without using the graphql library (could not reproduce):

```python
from typing import Generic, TypeVar, Union, Any, TypeGuard


class GraphQLType:
    pass


def is_type(type_: Any) -> bool:
    return isinstance(type_, GraphQLType)


GT = TypeVar("GT", bound=GraphQLType)


class GraphQLWrappingType(GraphQLType, Generic[GT]):
    """Base class for all GraphQL wrapping types"""

    of_type: GT

    def __init__(self, type_: GT) -> None:
        if not is_type(type_):
            raise TypeError(
                f"Can only create a wrapper for a GraphQLType, but got: {type_}."
            )
        self.of_type = type_


class GraphQLNamedType(GraphQLType):
    name: str

    def __init__(self, name: str) -> None:
        self.name = name


class GraphQLScalarType(GraphQLNamedType):
    pass


class GraphQLEnumType(GraphQLNamedType):
    pass


class GraphQLInputObjectType(GraphQLNamedType):
    pass


def is_enum_type(type_: Any) -> TypeGuard[GraphQLEnumType]:
    return isinstance(type_, GraphQLEnumType)


graphql_input_types = (GraphQLScalarType, GraphQLEnumType, GraphQLInputObjectType)

GraphQLInputType = Union[
    GraphQLScalarType, GraphQLEnumType, GraphQLInputObjectType, GraphQLWrappingType
]


def is_input_type(type_: Any) -> bool:
    return isinstance(type_, graphql_input_types) or (
        isinstance(type_, GraphQLWrappingType) and is_input_type(type_.of_type)
    )


class GraphQLInputField:
    type: GraphQLInputType

    def __init__(self, type_: GraphQLInputType) -> None:
        if not is_input_type(type_):
            raise TypeError("oops")

        self.type = type_


def parse_field(
    field: GraphQLInputField,
) -> None:
    if is_enum_type(field.type):
        print(field.type.name)
```

gaborbernat (approved):

src/datamodel_code_generator/parser/graphql.py
koxudaxi: @andrew-womeldorf 
`TypeGuard` may not work the case.
We should use `TypeIs` to avoid the problem. (we can create wrapper funciton.)
https://peps.python.org/pep-0742/

Issue #2321, opened 2025-02-11
Changing GitHub settings


gaborbernat:
I'd like to change a few settings here to make maintainership life easier:

- Allow only squash merging
- Enable: Always suggest updating pull request branches 
- Enable: Allow auto-merge Loading
- Enable: Automatically delete head branches Loading
- Enable: Limit how many branches and tags can be updated in a single push 

Delete the current old branch protection and replace it with a new ruleset, that:

- Bypass list: MaintainRole
- Target branches: Include default branch
- Enable:
  - Restrict deletions
  - Require linear history
  - Require a pull request before merging
     - Dismiss stale pull request approvals when new commits are pushed
     - Require approval of the most recent reviewable push
     - Require conversation resolution before merging
     - Allowed merge methods: squash
  - Require status checks to pass
     - enumerate the checks that are mandatory (let's not include the 3.8 ones) as want to put in a PR for removing 3.8 support soon, codecov also is not mandatory

Thanks! What options I did not enumerate should be off. Happy to discusss options which you do not agree with.

koxudaxi:
I am updating the settings and will come back

koxudaxi:
I have changed the settings.
I invoked the CI for test.
https://github.com/koxudaxi/datamodel-code-generator/pull/2322

PR #2324, opened 2025-02-14
Drop <3.9 support as run and generation target

gaborbernat:
This technically can be considered a bit breaking, as if someone passes
in `--python-target 3.6` that will now fail as we removed 3.6, but IMHO is the right approach.
Signed-off-by: Bernát Gábor <bgabor8@bloomberg.net>


gaborbernat:
@koxudaxi I wouldn't merge this without your approval but note that after https://github.com/koxudaxi/datamodel-code-generator/issues/2321#issue-2846147024 I should be able to bypass the branch protection, and I see I cannot 🤔 so something is not right there.

koxudaxi:
You're right, 3.6 is a good drop.
Because the implementation will also be more complicated.

We may get complaints about 3.8, but We will try it this time.

koxudaxi (approved):
we should bump the minor version to indicate the breaking changes of defualt behavior.

gaborbernat:
We're still doing zero-ver 😆 which has zero guarantees, but of course we should release it as 0.28.0 :D

gaborbernat:
> note that after [#2321 (comment)](https://github.com/koxudaxi/datamodel-code-generator/issues/2321#issue-2846147024) I should be able to bypass the branch protection, and I see I cannot 🤔

Can you check the ruleset? Would you like to need to approve every PR I make yourself, or you OK me landing commits without your approval?

koxudaxi:
image

OK, I have disabled the option

PR #2342, opened 2025-03-12
fix: Fix Union type parsing with recursive None removal

koxudaxi:


src/datamodel_code_generator/types.py
```diff
@@ -167,26 +167,71 @@ def _remove_none_from_type(type_: str, split_pattern: Pattern[str], delimiter: s
...
 
+        # Process each part of the union
+        parts = UNION_OPERATOR_PATTERN.split(type_)
+        processed_parts = []
+        for part in parts:
+            if part == NONE:
+                continue
+
+            # Check if this part contains a nested union
+            if " | " in part:
```
gaborbernat: Perhaps instead just update part and have a single append?

src/datamodel_code_generator/types.py
```diff
@@ -167,26 +167,71 @@ def _remove_none_from_type(type_: str, split_pattern: Pattern[str], delimiter: s
...
+            part = current_part[:-1].strip()
+            if part != NONE:
+                # Process nested unions recursively
+                if part.startswith(UNION_PREFIX):
+                    part = _remove_none_from_union(part, use_union_operator=False)
+                parts.append(part)
+            current_part = ""
+
+    if current_part:
+        part = current_part.strip()
```
gaborbernat: Should this be before the if?

gaborbernat (approved):

Issue #2336, opened 2025-03-04
[Regression] Invalid imports in generated models when referencing files with both minus (`-`) and dots (`.`)


ilovelinux:
**Describe the bug**
`datamodel-code-generator` doesn't support references to schemas with both minus (`-`) and dots (`.`) because file names are not sanitized. This lead to a crash because invalid python code is passed to black, which raises an exception.

This issue was (partially) fixed in:
- https://github.com/koxudaxi/datamodel-code-generator/issues/131 <- Match this issue
- https://github.com/koxudaxi/datamodel-code-generator/issues/1141 <- Partially match this issue (no dots)

May be related to:
- #2332 

**It works for references without any dot (`.`)**

For more details & reproducible example: https://github.com/ilovelinux/datamodel-code-generator-issue-2336

**To Reproduce**

Example schema:

`array-commons.schema.json`
```json
{
  "$schema": "https://json-schema.org/draft/2020-12/schema",
  "$id": "https://example.com/array-commons.schema.json",
  "title": "Commons",
  "description": "Commons objects",
  "$defs": {
    "defaultArray": {
      "type": "array",
      "minLength": 1,
      "maxLength": 100
    }
  }
}
```

`products.schema.json`
```json
{
  "$schema": "https://json-schema.org/draft/2020-12/schema",
  "$id": "https://example.com/products.schema.json",
  "title": "Products",
  "description": "The products in the catalog",
  "$ref": "array-commons.schema.json#/$defs/defaultArray",
}
```

Used commandline:
```
$ datamodel-codegen --input inputs/ --input-file-type jsonschema --output outputs/
```

**Expected behavior**

It should generate working code.

**Version:**
 - OS: Linux
 - Python version: 3.13
 - datamodel-code-generator version: 0.28.2

**Additional context**

For more details & reproducible example: https://github.com/ilovelinux/datamodel-code-generator-issue-2336

PR #2343, opened 2025-03-12
fix: Fix invalid imports for schema files with hyphens and dots

koxudaxi:
Fixes: https://github.com/koxudaxi/datamodel-code-generator/issues/2336


src/datamodel_code_generator/model/base.py
```diff
@@ -203,11 +204,19 @@ def get_template(template_file_path: Path) -> Template:
     return environment.get_template(template_file_path.name)
 
 
+def sanitize_module_name(name: str) -> str:
+    sanitized = re.sub(r"[^0-9a-zA-Z_]", "_", name)
+    if sanitized and sanitized[0].isdigit():
+        sanitized = "_" + sanitized
```
gaborbernat: I would use an f-string here.

tests/model/test_base.py
```diff
@@ -263,3 +265,37 @@ def test_data_field() -> None:
...
     field = DataModelFieldBase(name="a", data_type=DataType(is_list=True), required=False)
     assert field.type_hint == "Optional[List]"
+
+
+def test_sanitize_module_name() -> None:
+    assert sanitize_module_name("array-commons.schema") == "array_commons_schema"
+    assert sanitize_module_name("123filename") == "_123filename"
+    assert sanitize_module_name("normal_filename") == "normal_filename"
+    assert sanitize_module_name("file!name") == "file_name"
+    assert not sanitize_module_name("")
```
gaborbernat: This should be a parametrized test instead.

PR #2365, opened 2025-03-27
Extend parsing of Scientific Notation to JSON capabilites

maximilian-tech:
Solves Issues https://github.com/koxudaxi/datamodel-code-generator/issues/1955#issue-2290971596 and https://github.com/koxudaxi/datamodel-code-generator/issues/1952#issue-2287818185

It boils down to yaml vs. json floating point parsing. The SafeLoader of yaml has to be extended to allow for the same range of float notations like json.
[See here for more background](https://stackoverflow.com/questions/30458977/yaml-loads-5e-6-as-string-and-not-a-number).
I copied the SO-solution and it does work (for me)

This is needed to postprocess Fortran output as it often as a `e` (e.g. `1e-5`)for scientific notation, which yaml does not accept, but json does accept.
And because the json schema input is parsed through a yaml loader, this notation is not accepted. This PR extends the yaml loader to accept json notation of scientific data representation.


maximilian-tech:
@gaborbernat I rebased and added the CLI option --extend-yaml-scientifc-notation.

The current implementation uses a global variable, but this is the best solution I could come up with. Ideally the command line option is gone in the '1.0' release and with it this 'construct' so I think it is ok for now.

gaborbernat (changes requested):
Can you please add tests for the new features?

maximilian-tech:
I added one test.

The pre-commit-ci bot however 'fixes' the files in a broken state on code I haven't touched.

src/datamodel_code_generator/__init__.py
```diff
@@ -30,7 +30,7 @@
     PythonVersionMin,
 )
 from datamodel_code_generator.parser import DefaultPutDict, LiteralType
-from datamodel_code_generator.util import SafeLoader
+from datamodel_code_generator.util import SafeLoader, get_safeloader
```
koxudaxi, line 33: @maximilian-tech
Do you want to import `SafeLoader ` for the type-hint?
I think we should import `SafeLoader`  in the `if TYPE_CHECKING:` statement.

maximilian-tech: `SafeLoader` is the argument used in the `load_yaml` function. In `datamodel_code_generator.util` it is imported within the `TYPE_CHECKING` block. The `pre-commit.ci bot` does some code modifications beyond beautification, which changes the code as well.

I just provide the `get_safeloader` function and its wrapper, which acts on the global `g_extend_yaml_scientifc_notation` variable.

I wanted to perturb the original code as little as possible.

maximilian-tech:
What is needed to do progress here? I need this to parse json schema which contains Fortran style scientific notation ( e.g. `1e-5`) which is butchered by the yaml loader.

gaborbernat (changes requested):
CI Failing.

maximilian-tech:
Yes, because the pre-commit bot breaks the code.

gaborbernat:
> Yes, because the pre-commit bot breaks the code.

You'll need to address that 🤔

Issue #2364, opened 2025-03-26
Parameter name changing in Pydantic V2 model


ollyhensby:
**Describe the bug**
I think this is due to the changes in this PR: https://github.com/koxudaxi/datamodel-code-generator/pull/2355

We create the Pydantic model from JSON schema and then use that model but the name has appended `_1` on the parameter name so trying to set data to this field doesn't work anymore with the original parameter name. Is this appropriate behaviour to change the parameter name?

Is this so that the parameter name isn't the same as the enumeration class name? If so, maybe we keep the parameter name the same and use `Enum` as a suffix for the enumeration class instead of changing the parameter name?

**To Reproduce**
Running this:
```py
import pathlib
from datamodel_code_generator import DataModelType, InputFileType, generate
schema = {"title": "Test", "properties": {"Fruit": {"enum": ["apple", "banana"]}}}
output = pathlib.Path(__file__).parent / "test_model.py"
generate(
    json.dumps(schema),
    input_file_type=InputFileType.JsonSchema,
    input_filename="example.json",
    output=output,
    output_model_type=DataModelType.PydanticV2BaseModel,
    capitalise_enum_members=True,
)
```
produces
```py
# generated by datamodel-codegen:
#   filename:  example.json
#   timestamp: 2025-03-26T17:43:52+00:00

from __future__ import annotations

from enum import Enum
from typing import Optional

from pydantic import BaseModel, Field


class Fruit(Enum):
    APPLE = 'apple'
    BANANA = 'banana'


class Test(BaseModel):
    Fruit_1: Optional[Fruit] = Field(None, alias='Fruit')
```

Example schema:
```json
{
    "title": "Test",
    "properties": {
        "Fruit": {
            "enum": [
                "apple",
                "banana"
            ]
        }
    }
}
```

**Expected behaviour**
Maybe instead we could produce:
```py
# generated by datamodel-codegen:
#   filename:  example.json
#   timestamp: 2025-03-26T17:43:52+00:00

from __future__ import annotations

from enum import Enum
from typing import Optional

from pydantic import BaseModel, Field


class FruitEnum(Enum):
    APPLE = 'apple'
    BANANA = 'banana'


class Test(BaseModel):
    Fruit: Optional[FruitEnum] = Field(None, alias='Fruit')
```

**Version:**
 - OS: Ubuntu 22.04.5 LTS
 - Python version: 3.13.2
 - datamodel-code-generator version: 0.28.5

[...]

PR #2373, opened 2025-04-05
feat: add new option --parent-scoped-naming to avoid name collisions

gjcarneiro:
When you have object or enum definitions defined inline in operations, rather than referenced, you get the name defined globally in python.  If there are multiple versions with the same name, the name collision get solved by appending a number.

In some OpenAPI spec I am trying to use, which I didn't define, I am getting something like 12 "Status" enums, because of all the status fields.

This MR adds an option  --parent-scoped-naming so that the name of the generated python class is prefixed by  the name of the parent class from which it originated.

It kind of solves an issue similar to what is see in https://github.com/koxudaxi/datamodel-code-generator/issues/2364


guywaldman:
@koxudaxi This is a very useful flag to have. Could this be merged? 🙌

gaborbernat (changes requested):
Can you resolve the merge conflicts please?

gaborbernat (approved):

gjcarneiro:
I'm sorry if this option does not work correctly for your case.  I did not test this with any api specs that contain dots.  In my use cases, I always only have a single flat python module, without any dirs.

You can check around [this area](https://github.com/koxudaxi/datamodel-code-generator/blob/main/src/datamodel_code_generator/reference.py#L517-L528) what it is trying to do.  It simply concatenates the path of parent of the current model to the current model's path.  Perhaps it needs to be slightly tweaked when there is a `"."` (dot) in the path.

PR #2376, opened 2025-04-14
feat: save memory in pydantic2 by avoiding simple aliasing empty classes

gjcarneiro:
Insteaf of generating:

```py
class Bar(BaseModel):
   field1: ...

class Foo(Bar):
    pass
```

Wit this change we generate:

```py
class Bar(BaseModel):
   field1: ...

Foo = Bar
```

Thus saving one model class.  Because models in pydantic v2 consume a lot of memory.


gaborbernat (approved):

jrobichaud:
I found a workaround to disable it: https://github.com/koxudaxi/datamodel-code-generator/issues/2487#issuecomment-3228402186

PR #2394, opened 2025-04-29
Support `coerce_numbers_to_str` in `ConfigDict`

robotadam:
Support setting `coerce_numbers_to_str` in the Pydantic `ConfigDict` via `extra_template_data`.

https://docs.pydantic.dev/latest/api/config/#pydantic.config.ConfigDict.coerce_numbers_to_str

There are other ConfigDict values that you might consider adding as well, but this one was the blocker for us.


gaborbernat (changes requested):
Please add a test validating the change 👍

robotadam:
> Please add a test validating the change 👍

@gaborbernat I added it to the `extra_template_data` test; right now it only has one other entry from ConfigDict, but from what I can tell there was no reason to make an independent test — it'd be a lot of duplication if we have one for each config option. Let me know if you'd like it done another way!

gaborbernat (approved):

PR #2401, opened 2025-05-11
Handle JSON pointer escaped values

jrnold:
Handles JSON pointer escaping as defined in
[RFC-6901](https://datatracker.ietf.org/doc/html/rfc6901),
when looking up 

- `~0` is the escaped value of `/`
- `~1` is the escaped value of `~`
- URL encoded values are unescaped

Currently, reference lookup will not find the referenced location if the reference contains any escaped values.


jrnold:
Thanks for the package!

Issue #2388, opened 2025-04-23
Incorrect regex in generated pydantic model, when regex pattern is under oneOf/anyOf nested inside a property of a schema


pfaizian:
**Describe the bug**
If I include a regex pattern, under a oneOf/anyOf which is nested inside a property of a schema, the resulting pydantic mdoel has extra spaces after each comma in the regex. This doesn't happen without oneOf/anyOf, or when the oneOf/anyOf is at the top level of the schema.

**To Reproduce**

Example schema:
```json

openapi: "3.0.0"
components:
  schemas:
    Model1:
      description: Model1
      properties:
        strAttribute:
          type: string
          pattern: '^$|[^=,]+=(?:on|off)+(?:,[^=,]+=(?:on|off)+)*$'
    Model2:
      description: Model2
      properties:
        strOrIntAttribute:
          oneOf:
            - type: int
            - type: string
              pattern: '^$|[^=,]+=(?:on|off)+(?:,[^=,]+=(?:on|off)+)*$'
    StrOrIntAttribute:
      description: StrOrIntAttribute
      oneOf:
        - type: int
        - type: string
          pattern: '^$|[^=,]+=(?:on|off)+(?:,[^=,]+=(?:on|off)+)*$'

```

Used commandline:
```
$ datamodel-codegen --input api.yaml --output model.py
```

**Expected behavior**
For Model2, I expected to get the following pydantic model to be created:

```
class Model2(BaseModel):
    strOrIntAttribute: Optional[
        Union[Any, constr(regex=r'^$|[^=,]+=(?:on|off)+(?:,[^=,]+=(?:on|off)+)*$')]
    ] = None
```
But instead, I get this one (note the extra white spaces after each ",":
```
class Model2(BaseModel):
    strOrIntAttribute: Optional[
        Union[Any, constr(regex=r'^$|[^=, ]+=(?:on|off)+(?:, [^=, ]+=(?:on|off)+)*$')]
    ] = None
```

**Version:**
 - OS: macOS
 - Python version: 3.8.10
 - datamodel-code-generator version: 0.25.1

**Additional context**

kini:
Somewhat reproduced on current master with Python 3.13 on Linux:

```
[2] [kkini@kkini:/tmp]$ python -V
Python 3.13.1

[2] [kkini@kkini:/tmp]$ datamodel-codegen --version
0.30.1.dev1+gc0e19d92

[...]

PR #2404, opened 2025-05-23
fix regex rewrite bug

minomocca:
Solves issue #2388 

`_remove_none_from_union` splits `type_` at each comma and joins parts with `, `. This must not be done inside a constraint string (`constr`) as regex or pattern might change.

Added a counter for depth inside a constraint string for any non-escaped bracket.


gaborbernat (changes requested):
Please add tests validating this fix 👍

minomocca:
> Please add tests validating this fix 👍

`tests/test_types.py` has been updated, hope you are fine with this change (black changed a few more lines).

koxudaxi (approved):
LGTM

PR #2416, opened 2025-06-10
fix: documentation for --allow-extra-fields

### LINKED ISSUE #2409: Incorrect documentation for `--allow-extra-fields` flag
**Describe the bug**

The model customization flag `--allow-extra-fields` is documented as

```
--allow-extra-fields  Allow passing extra fields, if this flag is not passed, extra fields are forbidden.
```

I think that's incorrect. If the flag is passed, the models do indeed allow extra fields. But if the flag is not passed, extra fields may either be allowed, ignored, or forbidden, depending on the input.

**Example**

Given this schema

```json
{
  "type": "object",
  "properties": {
    "foo": {
      "type": "integer"
    }
  }
}
```

when I run

```
uv run datamodel-codegen --input-file-type jsonschema --input my_schema.json
```

it outputs

```
#   filename:  my_schema.json
#   timestamp: 2025-05-31T20:04:36+00:00

from typing import Optional

class Model(BaseModel):
    foo: Optional[int] = None
```

That model does not configure how extra data is treated, so uses the `pydantic` default (in both v1 and v2), which is to ignore it:

```
>>> m = Model(i_am_extra=1)
>>> m
Model(foo=None)
```

So: no `--allow-extra-fields` flag, but extra fields are ignored, not forbidden.

If your json schema specifies `"additionalProperties": true / false` then additional fields are allowed / forbidden, instead of ignored.

**Summary of actual behaviour**

For json schema:

| **--allow-extra-fields**<br>(when invoking codegen) | **"additionalProperties"**<br>(in jsonschema) | **extras**<br>(in generated model) |
|:---------------------------------------------------:|:---------------------------------------------:|:----------------------------------:|
|                       omitted                       |                    omitted                    |    pydantic default, i.e. ignore   |
|                       omitted                       |                      true                     |                allow               |
|                       omitted                       |                     false                     |               forbid               |
|                       present                       |                     \<any\>                     |                allow               |

both for output model type `pydantic.BaseModel` and `pydantic_v2.BaseModel`.

A quick test suggests the same is true for openapi specs.

I'm not familiar with GraphQL.

**Suggested change to the docs**

The actual behaviour is as I would expect. It's just the docs for `-allow-extra-fields` that I think is incorrect.

Perhaps update to something like this?

```
--allow-extra-fields  Allow passing extra fields. If this flag is not passed, the input schema controls how extra fields are handled, or else extra fields are ignored.
```

Or simply

```
--allow-extra-fields  Allow passing extra fields
```

leaving it to the user's common sense to guess what happens if the flag is not passed.

**Versions**
    OS: macOS sequoia 15.4.1
    Python: 3.13.2
    datamodel-code-generator: 0.30.1

Thanks for a great tool!

cosmo-grant:
## Summary

This tiny PR fixes the documentation for the model customization flag `--allow-extra-fields`.

Currently, that flag is documented as:

```
--allow-extra-fields Allow passing extra fields, if this flag is not passed, extra fields are forbidden.
```

That's incorrect. If the flag is passed, the models do indeed allow extra fields. But if the flag is not passed, extra fields may either be allowed, ignored, or forbidden, depending on the input.

So this PR updates the documentation to simply

```
--allow-extra-fields  Allow passing extra fields
```

I suggest that the flag's documentation need not describe what happens when the flag is omitted (namely, extra fields are treated however is specified in the input, defaulting to ignored), but leave that to the rest of the docs and the user's common sense. (My thought is: a flag's documentation need only describe how it _changes_ the typical behaviour, not the typical behaviour itself.)

## Deeper dive

For example, given this schema

```json
{
  "type": "object",
  "properties": {
    "foo": {
      "type": "integer"
    }
  }
}
```

when I run

```
uv run datamodel-codegen --input-file-type jsonschema --input my_schema.json
```

it outputs

```
#   filename:  my_schema.json
#   timestamp: 2025-05-31T20:04:36+00:00

from typing import Optional

class Model(BaseModel):
    foo: Optional[int] = None
```

That model does not configure how extra data is treated, so uses the `pydantic` default (in both v1 and v2), which is to ignore it:

```
>>> m = Model(i_am_extra=1)
>>> m
Model(foo=None)
```

More generally, here's the current actual behaviour:

| **--allow-extra-fields**<br>(when invoking codegen) | **"additionalProperties"**<br>(in input) | **extras**<br>(in generated model) |
|:---------------------------------------------------:|:---------------------------------------------:|:----------------------------------:|
|                       omitted                       |                    omitted                    |    pydantic default, i.e. ignore   |
|                       omitted                       |                      true                     |                allow               |
|                       omitted                       |                     false                     |               forbid               |
|                       present                       |                     \<any\>                     |                allow               |

both for output model type `pydantic.BaseModel` and `pydantic_v2.BaseModel`.

Fixes #2409


gaborbernat (changes requested):
Can you fix the merge conflicts?

Issue #1977, opened 2024-05-29
PydanticV2 allows extra fields _without_ `--allow-extra-fields` option


maximilian-tech:
**Describe the bug**
It should forbid `extra_fields` by default. PyDantic_v2 itself ignores `extra_fields` per default. So I expect `datamodel-code-gen` to explicitly set `allow_extra_fields=False` unless `--allow-extra-fields` is specified

**To Reproduce**

Example schema:
```json
{
  "type": "object",
  "properties": {
    "name": {
      "type": "string"
    },
    "age": {
      "type": "integer",
      "minimum": 0
    },
    "email": {
      "type": "string",
      "format": "email"
    },
    "address": {
      "type": "object",
      "properties": {
        "street": {
          "type": "string"
        },
        "city": {
          "type": "string"
        },
        "zipcode": {
          "type": "string",
          "pattern": "^[0-9]{5}(?:-[0-9]{4})?$"
        }
      },
      "required": ["street", "city", "zipcode"]
    },
    "phoneNumbers": {
      "type": "array",
      "items": {
        "type": "string",
        "pattern": "^[0-9]{10}$"
      }
    }
  },
  "required": ["name", "age", "email", "address"]
}

```

Used commandline:
```
$ datamodel-codegen --input test_json.json --output test_json.py --output-model-type pydantic_v2.BaseModel
```

**Expected behavior**
I want the following behavior by _default_
```
# generated by datamodel-codegen:
#   filename:  test_json.json
#   timestamp: 2024-05-29T06:01:45+00:00

from __future__ import annotations

from typing import List, Optional

from pydantic import BaseModel, ConfigDict, EmailStr, conint, constr


class Address(BaseModel):
    model_config = ConfigDict(
        extra='forbid',
    )
    street: str
    city: str
    zipcode: constr(pattern=r'^[0-9]{5}(?:-[0-9]{4})?$')


class Model(BaseModel):
    model_config = ConfigDict(
        extra='forbid',
    )
    name: str
    age: conint(ge=0)
    email: EmailStr
    address: Address
    phoneNumbers: Optional[List[constr(pattern=r'^[0-9]{10}$')]] = None

```
What I got is this:
```
# generated by datamodel-codegen:
#   filename:  test_json.json
#   timestamp: 2024-05-29T06:01:45+00:00

from __future__ import annotations

from typing import List, Optional

from pydantic import BaseModel, ConfigDict, EmailStr, conint, constr


[...]

Issue #1155, opened 2023-03-06
Can't get Extra.forbid using the command line 


Roobxyz:
Hello -- this forbid line (line 230) never gets hit and I think the default is `Extra.ignore` meaning there's no easy way to enforce strict type checking. The inline If statement just checks to see if `additionalProperties` or `allow_extra_fields` is set, but, if say addtionalProperties was set to `{'extra':'Extra.forbid'}` I think this line would set it back to `Extra.allow`.

https://github.com/koxudaxi/datamodel-code-generator/blob/97d4928782593efb1d439d91e7abbc70740b4ecd/datamodel_code_generator/model/pydantic/base_model.py#L224-L230

koxudaxi:
@Roobxyz 
We can set `additionalProperties: false` to set `extra = Extra.forbid `
Input:
https://github.com/koxudaxi/datamodel-code-generator/blob/7947f4dbf3efd7cef9209fa72660fb0a036f21c1/tests/data/openapi/additional_properties.yaml#L139-L149

Output:
https://github.com/koxudaxi/datamodel-code-generator/blob/7947f4dbf3efd7cef9209fa72660fb0a036f21c1/tests/data/expected/parser/openapi/openapi_parser_parse_additional_properties/with_import_format.py#L39-L44

git-malu:
```
components:
  schemas:
    AuthorizationSchema:
      $ref: "pap5.yaml#/components/schemas/AuthorizationSchema"
```


what if I'm referencing a schema, "AuthorizationSchema" for example, and I can't modify it directly since it's provided by others.

I have tried the method below, but it works for data model generator, but my editor complains "Schema validation: Property 'additionalProperties' is not allowed"
```
components:
  schemas:
    AuthorizationSchema:
      $ref: "pap5.yaml#/components/schemas/AuthorizationSchema"
      additionalProperties: false
```
Is this the right way to do it?
@koxudaxi

neelabhg:
Hi @koxudaxi, thank you for this excellent tool!
I have the same issue. For me, it's not practical to update the OpenAPI schema because it is provided by an external API vendor.

I can potentially ask them to add `additionalProperties: false` to all the schema models, but wouldn't it be better to default to `Extra.forbid` even when this is not specified?

koxudaxi:
@git-malu @neelabhg 
Sorry for the late reply.

> but wouldn't it be better to default to Extra.forbid even when this is not specified?

[...]

PR #949, opened 2022-12-20
Allow to pass extra fields using `--allow-extra-fields`

fgebhart:
Hi :wave:

Thanks for this great tool! Hope you are okay with adding this little addition.

This PR introduces support for a feature request described in https://github.com/koxudaxi/datamodel-code-generator/issues/785. As I found the code base super clean and the needed implementation rather straight forward, I'd like to contribute this change to your repo.

As it is my first time contributing to this tool, please pay attention to my naive implementation.  Please also check whether I might have forgotten or overlooked something, as my perspective might be limited to my narrow use-case. And of course I'm open to all kind of suggestions.

PR #2417, opened 2025-06-10
feat: add `--extra-fields` option to allow, forbid, or ignore extra fields

cosmo-grant:
## Summary

This PR adds an `--extra-fields` option, taking an argument of `allow`, `ignore` or `forbid`, to control how extra fields are handled in the generated models, overriding any `additionalProperties` settings in the input.

There is already an `--allow-extra-fields` option. But what if you want to forbid extra fields? Or want to ignore extra fields but the input has `"additionalProperties": false`? `--extra-fields` addresses these use cases.

`--extra-fields` includes the functionality of `--allow-extra-fields`, since it seemed sensible to me to have _one_ option to allow, ignore, or forbid, instead of one option to allow and another option to ignore or forbid. So this PR deprecates `--allow-extra-fields` in favour of `--extra-fields=allow`.

It's backwards compatible. In particular, it doesn't remove `--allow-extra-fields` and the default behaviour if `--extra-fields` is not passed is unchanged.

The implementation follows quite closely #949, for adding `--allow-extra-fields`.

## Example

Given the schema

```json
{
  "title": "Test",
  "type": "object",
  "required": [
    "foo"
  ],
  "properties": {
    "foo": {
      "type": "object",
      "properties": {
        "x": {
          "type": "integer"
        }
      },
      "additionalProperties": true
    },
    "bar": {
      "type": "object",
      "properties": {
        "y": {
          "type": "integer"
        }
      },
      "additionalProperties": false
    },
    "baz": {
      "type": "object",
      "properties": {
        "z": {
          "type": "integer"
        }
      }
    }
  },
  "additionalProperties": false
}
```

when I run

```
uv run datamodel-codegen --input-file-type jsonschema --input tests/data/jsonschema/extra_fields_ignore.json --extra-fields=ignore
```

it outputs

```python
#   filename:  extra_fields_ignore.json
#   timestamp: 2025-06-10T21:10:42+00:00

from typing import Optional

from pydantic import BaseModel, Extra


class Foo(BaseModel):
    class Config:
        extra = Extra.ignore

    x: Optional[int] = None


class Bar(BaseModel):
    class Config:
        extra = Extra.ignore

    y: Optional[int] = None


class Baz(BaseModel):
    class Config:
        extra = Extra.ignore

    z: Optional[int] = None


class Test(BaseModel):
    class Config:
        extra = Extra.ignore

    foo: Foo
    bar: Optional[Bar] = None
    baz: Optional[Baz] = None
```

i.e. extra fields are explicitly ignored everywhere in the generated models, overriding the `"additionalProperties"` settings in the input.

This is similar to how the existing flag `--allow-extra-fields` works, but lets you ignore or forbid, as well as allow.

## Use case

Here's one use case, for example.

Suppose your jsonschema includes `"additionalProperties": false`, because you use it in a cloud service's api gateway to automatically validate api requests and do want to fail payloads with additional properties.

Even so, you might want to generate a model that ignores extra fields instead. (As things stand, the model would forbid them by default or allow them if you pass `--allow-extra-fields`.) For example, the data to instantiate the model might arrive in your app in three dictionaries, along with extra fields. It would be convenient to instantiate the model via `m = MyModel(**d1, **d2, **d3)`, letting pydantic filter out the extra fields.

If the model _forbids_ extra fields, then trying to instantiate the model like that throws a validation error.  If the model _allows_ extra fields, then the extra fields become attributes of the instance, e.g. they show up when you `m.model_dump()`. But if you can make the model _ignore_ extra fields, via `--extra-fields=ignore`, then it all works out: no validation error, and no unwanted fields, without having to filter fields yourself.

## Related open issues

In #1977, the OP wanted extra fields to be forbidden by default, but in a follow-up comment wrote:

> But for me it would be ok, if the documentation would describe the implementation, and an additional flag would be introduced - keeping the status quo of not interfering with 'default' values.

which I believe is addressed by this PR.

I believe this PR also addresses the main point of #1155.


gaborbernat (approved):

cosmo-grant:
I guess the gif means you'd ideally like this PR done a bit differently? I'm happy to work on that if you let me know what you have in mind.

Thanks for a great tool!

Issue #2357, opened 2025-03-24
feature: adding parameter to @dataclass()


puttemanss:
**Is your feature request related to a problem? Please describe.**
No real problem

**Describe the solution you'd like**
be able to parameter to dtaclass for example @dataclass(slots=True)

**Describe alternatives you've considered**
Nothing

**Additional context**
https://docs.python.org/3/library/dataclasses.html
@dataclasses.dataclass(*, init=True, repr=True, eq=True, order=False, unsafe_hash=False, frozen=False, match_args=True, kw_only=False, slots=False, weakref_slot=False)

PR #2437, opened 2025-06-30
add dataclass arguments

ICEPower420:
Resolves feature: adding parameter to @dataclass() #2357
Added argument dataclass_arguments takes precedence over --keyword-only and --frozen-dataclasses

### DIFF
--- a/src/datamodel_code_generator/__init__.py
+++ b/src/datamodel_code_generator/__init__.py
@@ -290,6 +290,7 @@ def generate(  # noqa: PLR0912, PLR0913, PLR0914, PLR0915
     no_alias: bool = False,
     formatters: list[Formatter] = DEFAULT_FORMATTERS,
     parent_scoped_naming: bool = False,
+    dataclass_arguments: dict[str, Any] | None = None,
 ) -> None:
     remote_text_cache: DefaultPutDict[str, str] = DefaultPutDict()
     if isinstance(input_, str):
@@ -304,6 +305,11 @@ def generate(  # noqa: PLR0912, PLR0913, PLR0914, PLR0915
     else:
         input_text = None
 
+    dataclass_arguments = dict(dataclass_arguments or {})
+    dataclass_arguments.setdefault("frozen", frozen_dataclasses)
+    dataclass_arguments.setdefault("kw_only", keyword_only)
+
+    
     if isinstance(input_, Path) and not input_.is_absolute():
         input_ = input_.expanduser().resolve()
     if input_file_type == InputFileType.Auto:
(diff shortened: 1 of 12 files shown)


gaborbernat (changes requested):
CI is failing sadly.

gaborbernat (approved):

gaborbernat (changes requested):
You'll need to fix the readme failure:

```diff
diff --git a/README.md b/README.md
index e7d4b968..5347439b 100644
--- a/README.md
+++ b/README.md
@@ -495,7 +495,7 @@ Template customization:
                         A file with kwargs for custom formatters.
   --custom-template-dir CUSTOM_TEMPLATE_DIR
                         Custom template directory
-  --encoding ENCODING   The encoding of input and output (default: cp1252)
+  --encoding ENCODING   The encoding of input and output (default: utf-8)
   --extra-template-data EXTRA_TEMPLATE_DATA
                         Extra template data
   --use-double-quotes   Model generated with double quotes. Single quotes or your black
diff --git a/docs/index.md b/docs/index.md
index 0cc303ab..478dd24a 100644
--- a/docs/index.md
+++ b/docs/index.md
@@ -487,7 +487,7 @@ Template customization:
                         A file with kwargs for custom formatters.
   --custom-template-dir CUSTOM_TEMPLATE_DIR
                         Custom template directory
-  --encoding ENCODING   The encoding of input and output (default: cp1252)
+  --encoding ENCODING   The encoding of input and output (default: utf-8)
   --extra-template-data EXTRA_TEMPLATE_DATA
                         Extra template data
   --use-double-quotes   Model generated with double quotes. Single quotes or your black
```

Otherwise looks good.

gaborbernat (approved):

docs/index.md
```diff
@@ -483,7 +487,7 @@ Template customization:
                         A file with kwargs for custom formatters.
   --custom-template-dir CUSTOM_TEMPLATE_DIR
                         Custom template directory
-  --encoding ENCODING   The encoding of input and output (default: utf-8)
```
ilovelinux, line 488: Why has this line has changed? I can't find any edit related to default file encoding

gaborbernat: Yeah the CI is breaking because of this, my guess is that the readme generator behaves differently on Windows and UNIX, and the PR creator ran it on Windows causing this change.

gaborbernat (changes requested):

docs/index.md
ilovelinux: OT: I wonder if we should set a default encoding instead of relying on the system default/

Or we could specify that it's the SYSTEM default (and it's not enforced by the tool)

docs/index.md
```diff
@@ -483,7 +487,7 @@ Template customization:
...
+  --encoding ENCODING   The encoding of input and output (default: cp1252)
```
ilovelinux:
```suggestion
  --encoding ENCODING   The encoding of input and output (default: utf-8)
```

ilovelinux (dismissed):

PR #2445, opened 2025-07-07
Allow including path parameters in generated models

MrLoh:
This PR adds a new `--include-path-parameters` CLI flag that enables the generation of parameter models containing both path and query parameters when using `--openapi-scopes parameters`.

### Changes

- Added `--include-path-parameters` flag to CLI arguments
- Updated parameter model generation to include path parameters when flag is enabled
- Changed model suffix from `ParametersQuery` to `Parameters` when path parameters are included
- Added test coverage for the new functionality
- Added exception on name conflicts between path and query parameters.

### Rationale

I need this to validate the complete parameter set for an endpoint (e.g., `GET /users/{user_id}?include=profile&format=json` generates a single `Parameters` model containing `user_id`, `include`, and `format` fields) for use in SDKs.

The flag is opt-in to maintain backward compatibility.

### DIFF
--- a/tests/parser/test_openapi.py
+++ b/tests/parser/test_openapi.py
@@ -696,7 +696,7 @@ def test_parse_all_parameters_duplicate_names_exception() -> None:
         ParameterObject.parse_obj({"name": "duplicate_param", "in": "query", "schema": {"type": "integer"}}),
     ]
 
-    with pytest.raises(Exception) as exc_info:
+    with pytest.raises(Exception) as exc_info:  # noqa: PT011
         parser.parse_all_parameters("TestModel", parameters, ["test", "path"])
 
     assert "Parameter name 'duplicate_param' is used more than once." in str(exc_info.value)


gaborbernat (approved):

MrLoh:
@gaborbernat thanks for reviewing and approving. I fixed the black formatting issues in the tests and verified that they are passing locally, can you approve the CI runs and merge if they pass.

src/datamodel_code_generator/parser/openapi.py
```diff
@@ -421,8 +423,17 @@ def parse_all_parameters(
...
+            if (
+                not parameter_name
+                or parameter.in_ not in {ParameterLocation.query, ParameterLocation.path}
+                or (parameter.in_ == ParameterLocation.path and not self.include_path_parameters)
+            ):
                 continue
+
+            if any(field.original_name == parameter_name for field in fields):
+                msg = f"Parameter name '{parameter_name}' is used more than once."
+                raise Exception(msg)  # noqa: TRY002
```
MrLoh, line 435: @gaborbernat CI complains that I'm missing code coverage for these lines. It seems a little silly to add a test for this sanity check to me. But happy to do so if that's required.

MrLoh:
I added a test for the uncovered lines and updated the readme and docs with the script

gaborbernat:
You need to fix the linter errors now 😊

Issue #2451, opened 2025-07-21
Crash when providing JSON schema via `<stdin>`


adamrimon:
When providing a JSON schema via `<stdin>` rather than `--input`, reference handling may crash.
This happens because the absent filename (or `path_parts`) is replaced with a `"#"` (in jsonschema.py) that collides with `ID_PATTERN` pattern matching (in reference.py).

**To Reproduce**

Example schema:
```json

{
  "$schema" : "http://json-schema.org/draft-2019-09/schema",
  "type" : "object",
  "oneOf" : [ {
    "$ref" : "#/definitions/test"
  }],
  "definitions": {
    "test": {
      "type": "object",
      "properties": {
        "name": {
          "type": "string"
        }
      }
    }
  }
}

```

Used commandline:
```
$ datamodel-codegen --input-file-type jsonschema < schema.json
```

**Expected behavior**
This schema should just be processed correctly, like when runnning:
```
$ datamodel-codegen --input-file-type jsonschema --input schema.json
```

**Version:**
 - OS: Mac OS X
 - Python version: 3.12
 - datamodel-code-generator version: 0.31.2
 
**Solution Suggestions**
The easiest fix would be to change the _default_ `path_parts` used for `<stdin>` with `"."`, for example.

PR #2452, opened 2025-07-21
fixed: empty path (when input comes from <stdin>) is now replaced with a "." instead of a "#" (#2451)

adamrimon:


gaborbernat (changes requested):
Please add tests 👍

gaborbernat:
Seems this PR stalled, I'll close it now, we can re-open it if you pick it up again, just ping me.

Issue #2314, opened 2025-02-09
TypedDict Generation not covering `const` `required`


Mazyod:
**Describe the bug**
Generating `TypedDict` definitions of a schema with `const` and `required` yields `NotRequired[str]`.

**To Reproduce**

Example schema:
```json
{
  "$schema": "http://json-schema.org/draft-07/schema#",
  "definitions": {
    "AndType": {
      "additionalProperties": false,
      "description": "...",
      "properties": {
        "kind": {
          "const": "and",
          "type": "string"
        }
      },
      "required": [
        "kind"
      ],
      "type": "object"
    }
  }
}
```

Used commandline:
```
$ datamodel-codegen \
  --input ./schema.json \
  --output ./schema.py \
  --output-model-type "typing.TypedDict" \
  --target-python-version "3.11"
```

Generated code:
```python
# generated by datamodel-codegen:
#   filename:  schema.json
#   timestamp: 2025-02-09T17:42:53+00:00

from __future__ import annotations

from typing import Any, NotRequired, TypedDict

Model = Any


class AndType(TypedDict):
    kind: NotRequired[str]
```

**Expected behavior**
Generate `kind: Literal["and"]` since it is a `required` `const` property.

**Version:**
 - OS: macOS
 - Python version: 3.13.1
 - datamodel-code-generator version: 0.27.2

PR #2475, opened 2025-08-11
TypedDict: Add support for const fields using Literal

erandagan:
See: https://github.com/koxudaxi/datamodel-code-generator/issues/2314


gaborbernat (dismissed):

gaborbernat:
Let me see https://github.com/koxudaxi/datamodel-code-generator/pull/2479

gaborbernat:
Hmm passed on main, so somehow must be related to your changes 🤔

ysndr:
Hej, may I ask whats the status of this, now that tests appear to pass?
Also I'd love to give the same treatment to the dataclass implementation.
In fact it looks like its works with the exact same diff.
If you agree, would you prefer this rolled into one PR or separately?
@erandagan @koxudaxi

PR #2504, opened 2025-09-26
fix: set not nullable not required Struct fields to UnsetType

CharString:
Not nullable not required object parameters are not `typing.Optional`. `msgspec` has an `UnsetType`, which has the same semantics as `typing.NotRequired`.


koxudaxi:
@CharString 
Could plesae fix the unittest errors?

Issue #1285, opened 2023-04-28
Add an option to generate TypeAlias for literals when using the `--enum-field-as-literal`


mmohaveri:
**Is your feature request related to a problem? Please describe.**

Currently `datamodel-code-generator` has an option to parse enums as python's string literal (e.g. Literal["A", "B"]), this is great and I'm using it a lot, but the problem is if I want to pass these literal around in my code, I have to define their type once again. e.g:

```ptyhon
MyLiterals: TypeAlias = Literal["A", "B"]

def my_func(l: MyLiterals): ...
```

This is in contrast with all other types and structures generated by `datamodel-code-generator`. For example if had chosen to use python enums, I would have a `MyLiteral` enum in the generated code and could have use it in the function's type hint.

**Describe the solution you'd like**
I like to have an additional option in the CLI to tell  `datamodel-code-generator` to create these literal definitions automatically.

**Describe alternatives you've considered**

Currently I just copy and paste the literals generated by  `datamodel-code-generator` into a separate file and re-define them, which can be a little non-DRY.

mmohaveri:
@koxudaxi Is this something that you agree with and plan to do? If so I may be able to send a PR for it if you haven't planed and worked on it yet

koxudaxi:
@mmohaveri 
Thank you for creating the issue.
I want the feature too.
But, I don't have the time :(
I'm happy if you create the PR.

mmohaveri:
@koxudaxi  Sure, I'll see what I can do and send a PR

Issue #2018, opened 2024-06-28
Generate Annotated types instead of root types for string constraints


RonnyPfannschmidt:
**Is your feature request related to a problem? Please describe.**

given a schema like 
```yaml
components:
  schemas:
    HereBeDragons:
      type: string
      pattern: here
    PassMe:
      type: object
      properties:
        mangled_string:
          anyOf:
            - $ref: '#/components/schemas/PettyNumber'
            - $ref: '#/components/schemas/HereBeDragons'
    PettyNumber:
      type: string
      pattern: '[0-9]{2,8}'
  
```
multiple root types are created and in turn create object indirection

```python

from pydantic import BaseModel, Field, RootModel, constr


class HereBeDragons(RootModel[constr(pattern=r"here")]):
    root: constr(pattern=r"here")


class MoreValues(BaseModel):
    name: str
    age: int | None = None


class PettyNumber(RootModel[constr(pattern=r"[0-9]{2,8}")]):
    root: constr(pattern=r"[0-9]{2,8}")

class PassMe(BaseModel):
    mangled_string: PettyNumber | HereBeDragons | None = None
  

```
**Describe the solution you'd like**

decare the types more directly like

```python
# using python 3.12+ syntax

from pydantic import BaseModel, Field, StringConstraints


type HereBeDragons = Annotated[str, StringConstraints(pattern=r"here")]
type PrettyNumber = Annotated[str, StringConstraints(pattern=r"[0-9]{2,8}")]


class PassMe(BaseModel):
    mangled_string: PettyNumber | HereBeDragons | None = None
```

**Additional context**

RonnyPfannschmidt:
At first glance it seems like that is the case

shurshilov:
I think --use-type-alias  its not same.
need like --use-annotated but its already exist but it doesn't work as expected.. 
The problem is that the RootModel is generally redundant and slower. It is also not hashable and such types cannot be used in a python set(), this leads to errors. In sum, the RootModel generation is essentially a bug. You should use Annotated out of the box. You can see above how many likes the request has garnered. type-aliase what you are suggesting is not what it is though the developers) Please take more time for this.

PR #861, opened 2022-09-28
fix: template handling of extra template data

auphofBSF:
see issue #860 , this commit includes
- fixes to template,
- extra test case and supporting, data and outputs

NOT INCLUDED
- issue with Field description string delimiting NOT FIXED HERE 
- documentation updates

PR #2505, opened 2025-09-27
Add --use-type-alias flag to generate TypeAlias instead of root models

### LINKED ISSUE #2427: Use annotated type instead of the RootModel
Currently, the schema generator creates a lot of custom types, which are hard to use:
```
class Total(RootModel[int]):
    root: Annotated[int, Field(ge=0, title="Total")]


class Page(RootModel[int]):
    root: Annotated[int, Field(ge=1, title="Page")]
```
It doesn't work with `int` when you have `validate_call` pydantica validation, and it is properly stringified using `f-strings` (it became `root=42` instead of `42`)

I would much rather prefer to see just an annotated field instead:
```
Total = Annotated[int, Field(ge=0, title="Total")]
Page = Annotated[int, Field(ge=1, title="Page")]
```
So, we have the same level of validation, but we use a built-in type for that

**Additional context**
I updated a third-party library (Beanie), which changed the type from `str` to custom `PydanticObjectId`, so I have to fix existing code in multiple places to be able to use auto-generated models still

butvinm:
Closes #2427

This adds a new CLI flag --use-type-alias that generates Python type aliases using the TypeAlias syntax instead of root models for simple types like unions and arrays.

Example output:
```python
MySchema: TypeAlias = Annotated[Union[str, int, bool], Field(..., title='MySchema')]
```
Instead of:
```python
class MySchema(RootModel[Union[str, int, bool]]):
    root: Union[str, int, bool] = Field(..., title='MySchema')
```


src/datamodel_code_generator/arguments.py
```diff
@@ -305,6 +305,10 @@ def start_section(self, heading: str | None) -> None:
     action="store_true",
     default=None,
 )
+typing_options.add_argument(
+    "--use-type-alias",
+    help="Use TypeAlias instead of root models",
+)
```
ilovelinux, line 314: This should fix the error in the test and allow them to run correctly.

```suggestion
    action="store_true",
    default=None
)
```

butvinm: thanks, fixed

tests/data/expected/main/graphql/type_alias.py
ilovelinux, line 1: Why are there descriptions of the type aliases? Have they been automatically generated by the library?

They are incorrect. E.g. int in python can be larger than 32bit and there are no additional constraints in the definition of `Int` about that

butvinm: For GraphQL the library generates default type aliases for Int, String and similar types, you can find it in other tests (https://github.com/koxudaxi/datamodel-code-generator/blob/main/tests/data/expected/main/graphql/union.py), so I suppose this issue is not in the scope of this PR

butvinm: I will fix how we handle TypeAlias for different python versions, thanks for pointing out

ilovelinux: @butvinm consider using `target_version` cli property value

butvinm: I've updated the handling of `TypeAlias` across different Python versions:
- Python 3.9: `typing_extensions.TypeAlias`
- Python 3.10–3.11: `typing.TypeAlias`
- Python 3.12+: `type`

Additionally, I replaced the dummy `RootModel` with `TypeAlias` for `TypedDict`, `msgspec`, and `dataclass` outputs. This change generates proper type annotations instead of plain assignments. This is slightly out of the scope of this PR, but I decided to implement it for consistency.

I've added details about new behavior to the issue description

src/datamodel_code_generator/model/__init__.py
ilovelinux, line 1: :warning: We still need `RootModel` because type aliases doesn't support field-specific metadata!

> Pydantic will not understand field-specific metadata inside named aliases.

See https://docs.pydantic.dev/latest/concepts/types/#named-type-aliases

src/datamodel_code_generator/model/__init__.py
```diff
@@ -22,62 +22,97 @@ class DataModelSet(NamedTuple):
...
+    if target_python_version.has_type_statement:
+        type_alias_class = type_alias.TypeStatement
+        scalar_class = scalar.DataTypeScalarTypeStatement
+        union_class = union.DataTypeUnionTypeStatement
+    elif target_python_version.has_type_alias:
+        type_alias_class = type_alias.TypeAlias
+        scalar_class = scalar.DataTypeScalar
+        union_class = union.DataTypeUnion
+    else:
+        type_alias_class = type_alias.TypeAliasBackport
```
ilovelinux: :warning: Pydantic doesn't know the TypeAlias variable name. We need to use `TypeAliasType` instead. This way it will create the same object created using `type` keyword.

See:
- https://docs.pydantic.dev/latest/concepts/types/#named-type-aliases
- https://peps.python.org/pep-0695/#runtime-type-alias-class

tests/data/expected/main/jsonschema/type_alias_with_field_description_py312.py
ilovelinux, line 1: I see there are tests for `py39`, `py310-311`, and `py312+`. Since the default target version is `py39` I'd suggest to rename those tests as follows:
|Actual name|New name|
|---|---|
|`typed_dict_py39`|`typed_dict`|
|`typed_dict`|`typed_dict_py310`|

Using `typed_dict` with the output given by datamodel-codegen when the target version isn't specified

src/datamodel_code_generator/model/__init__.py
butvinm: Maybe I don't get something from the Pydantic documentation, but I expected `TypeAlias` to cause an error and `RootModel` to work when using `default`, but I ran into the opposite behavior.

**With RootModel:**
```python
from typing import Annotated
from pydantic import BaseModel, Field, RootModel

MyAlias = RootModel[Annotated[int, Field(default=1)]]
    
class Model(BaseModel):
    x: MyAlias

print(Model())
print(Model.model_json_schema())
```
**Output:**
```
pydantic_core._pydantic_core.ValidationError: 1 validation error for Model
x
  Field required [type=missing, input_value={}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.10/v/missing
```

**With unnamed TypeAlias:**
```python
from typing import Annotated, TypeAlias

MyAlias: TypeAlias = Annotated[int, Field(default=1)]

class Model(BaseModel):
    x: MyAlias

print(Model())
print(Model.model_json_schema())
```
**Output (json schema is incomplete):**
```
x=1
{'properties': {'x': {'default': 1, 'title': 'X', 'type': 'integer'}}, 'title': 'Model', 'type': 'object'}
```

**With named type alias:**
```python
from typing import Annotated, TypeAlias

type MyAlias = Annotated[int, Field(default=1)]

class Model(BaseModel):
    x: MyAlias

print(Model())
print(Model.model_json_schema())
```
**Output:**
```
x=1
{'$defs': {'MyAlias': {'default': 1, 'type': 'integer'}}, 'properties': {'x': {'$ref': '#/$defs/MyAlias'}}, 'title': 'Model', 'type': 'object'}
```

While both named type alias and RootModel fails on `alias`:
```python
from typing import Annotated

from pydantic import BaseModel, ConfigDict, Field, RootModel


MyAlias = RootModel[Annotated[int, Field(default=1, alias="y")]]

class Model(BaseModel):
    model_config = ConfigDict(extra="forbid")
    x: MyAlias


print(Model(y=10))
print(Model.model_json_schema())
```

**Output:**
```
x
  Field required [type=missing, input_value={'y': 10}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.10/v/missing
y
  Extra inputs are not permitted [type=extra_forbidden, input_value=10, input_type=int]
    For further information visit https://errors.pydantic.dev/2.10/v/extra_forbidden
```

As I understand, both RootModel and named type alias does not support field-specific metadata.

This behavior contradicts the documentation and does not show how `RootModel` is more helpful then named type alias in our scenario. 

One possible solution I see is to warn the user when field-specific metadata is defined in the schema that we would generate as either RootModel or TypeAlias, but that is for future consideration. What do you think? Could you clarify me if I got something wrong

ilovelinux: It may be an issue that is already in the codebase with RootModel but we didn't notice that.

I remember that when RootModel is used, it's inherited by a class. I don't know if it changes the behavior.

Also, things like default values may be put one level up. E.g. the first example may be:

```python
MyAlias = RootModel[int]
    
class Model(BaseModel):
    x: MyAlias = 1
```

So maybe the generator is working right and this is not an issue.

We should investigate more. I'll try to write down a checkpoint list of things to verify.

Maybe @koxudaxi or @gaborbernat may help us to crack the case.

(OT: Please add https://github.com/koxudaxi/datamodel-code-generator/issues/2018 too to the PR description as the author confirmed it should solve that one too! :slightly_smiling_face: )

butvinm: The inheritance form of RootModel behaves the same. 

My assumption is that both RootModel and type aliases are limited and it is not a generation problem, but we definitely should investigate how they work in combination with field properties, e.g., when RootModel is referenced as a field type and this field has its own constraints and default values. I will research too and share my insights.

tests/data/jsonschema/type_alias.json
```diff
@@ -0,0 +1,34 @@
...
+      "type": "object",
+      "properties": {
+        "simple_field": {"$ref": "#/definitions/SimpleString"},
+        "union_field": {"$ref": "#/definitions/UnionType"},
+        "array_field": {"$ref": "#/definitions/ArrayType"},
+        "annotated_field": {"$ref": "#/definitions/AnnotatedType"}
+      }
+    }
+  }
+}
```
ilovelinux, line 35: Adding `description` to the definitions of `SimpleString`, `UnionType`, and `AnnotatedType`, and using `--use-field-description`, generates code with missing field description.

```suggestion
{
  "$schema": "http://json-schema.org/draft-07/schema#",
  "definitions": {
    "SimpleString": {
      "description": "An annotated type alias SimpleString",
      "type": "string"
    },
    "UnionType": {
      "description": "An annotated type alias UnionType",
      "anyOf": [
        {"type": "string"},
        {"type": "integer"}
      ]
    },
    "ArrayType": {
      "description": "An annotated type alias ArrayType",
      "type": "array",
      "items": {"type": "string"}
    },
    "AnnotatedType": {
      "title": "MyAnnotatedType",
      "description": "An annotated type alias AnnotatedType",
      "anyOf": [
        {"type": "string"},
        {"type": "boolean"}
      ]
    },
    "ModelWithTypeAliasField": {
      "type": "object",
      "properties": {
        "simple_field": {"$ref": "#/definitions/SimpleString"},
        "union_field": {"$ref": "#/definitions/UnionType"},
        "array_field": {"$ref": "#/definitions/ArrayType"},
        "annotated_field": {"$ref": "#/definitions/AnnotatedType"}
      }
    }
  }
}
```

Output WITHOUT `--use-type-alias`:
```py
# $ datamodel-codegen --input test.json --output test.py --use-field-description
#
#   filename:  test.json
#   timestamp: 2025-10-01T21:19:55+00:00

from typing import Any, List, Optional, Union

class Model(BaseModel):
    __root__: Any


class SimpleString(BaseModel):
    __root__: str
    """
    An annotated type alias SimpleString
    """


class UnionType(BaseModel):
    __root__: Union[str, int]
    """
    An annotated type alias UnionType
    """


class ArrayType(BaseModel):
    __root__: List[str]
    """
    An annotated type alias ArrayType
    """


class AnnotatedType(BaseModel):
    __root__: Union[str, bool] = Field(..., title='MyAnnotatedType')
    """
    An annotated type alias AnnotatedType
    """


class ModelWithTypeAliasField(BaseModel):
    simple_field: Optional[SimpleString] = None
    union_field: Optional[UnionType] = None
    array_field: Optional[ArrayType] = None
    annotated_field: Optional[AnnotatedType] = None
```

Output WITH `--use-type-alias`:
```py
# $ datamodel-codegen --input test.json --output test.py --use-type-alias --use-field-description
#   filename:  test.json
#   timestamp: 2025-10-01T21:21:42+00:00

from typing import Annotated, Any, List, Optional, Union

from typing_extensions import TypeAlias

Model: TypeAlias = Any


SimpleString: TypeAlias = str


UnionType: TypeAlias = Union[str, int]


ArrayType: TypeAlias = List[str]


AnnotatedType: TypeAlias = Annotated[
    Union[str, bool], Field(..., title='MyAnnotatedType')
]


class ModelWithTypeAliasField(BaseModel):
    simple_field: Optional[SimpleString] = None
    union_field: Optional[UnionType] = None
    array_field: Optional[ArrayType] = None
    annotated_field: Optional[AnnotatedType] = None
```

ilovelinux:
This PR should also fix https://github.com/koxudaxi/datamodel-code-generator/issues/1285

src/datamodel_code_generator/model/__init__.py
butvinm: I replaced `TypeAlias` with `typing_extensions.TypeAliasType` for python 3.9 - 3.11

tests/data/expected/main/jsonschema/type_alias_with_field_description_py312.py
butvinm: Tests for python 3.9 and 3.10 were converged, so there is no need for renaming anymore

tests/data/jsonschema/type_alias.json
butvinm: I have fixed this issue and added corresponding test cases

butvinm:
@ilovelinux  I tried this version on my work codebase, and the generated models seem fine and pass tests that do not rely on `.root` access. However, I found another limitation of type aliases: they are missing `model_config`. In such situations, you must either lift the config to the model where the alias is used as a field type (e.g., set `regex_engine` on the base class) or be stuck with `RootModels`.

butvinm:
@koxudaxi What should I do with `tests/data/expected/main/graphql/github_api.py` - update with new datamode-code-generator version or remove? It is never referenced in the tests

koxudaxi:
@butvinm 
If you are not using it for testing, please delete it.a

src/datamodel_code_generator/model/template/TypeAliasType.jinja2
```diff
@@ -0,0 +1,20 @@
...
+{%- if _field.annotated -%}
+{{ _field.annotated }}
+{%- elif _field.field -%}
+Annotated[{{ _field.type_hint }}, {{ _field.field }}]
+{%- else -%}
+{{ _field.type_hint }}
+{%- endif -%}
+{%- endmacro -%}
+
+{{ class_name }} = TypeAliasType("{{ class_name }}", {{ get_type_annotation(fields[0]) }}){% if comment is defined %}  # {{ comment }}{% endif %}
```
ilovelinux, line 11: What's `{{ comment }}`? Could you provide me an example? I'm not familiar with it.

butvinm: I copied this from other templates.

I investigated that it was added in #861 as part of the work on extra template data support, but I don't see any reason for this field to be in all default templates, in my opinion. However, removing it is outside the scope of this PR.

butvinm: Generating a `type alias` instead of a `RootModel` for the main model is the intended behavior.

For example, in the JSON Schemas at my work, we have some models defined like this as dedicated JSON Schema files:

```json
{
  "title": "csvErrorType",
  "type": "string"
}
```

This is essentially a type alias and I expect it to generate `type alias` if `--use-type-alias` flag is enabled.

What is your use case for having a `RootModel` for the entrypoint model but `type aliases` for others?

ilovelinux: @butvinm we inherit the generated models to add more validation/customization layers. Most of the models are complex enough to not be collapsed by `--use-type-alias`, but we have a few generic models that are just `dict[str, str]` or similar, and they become totally collapsed. The behavior is theoretically correct, but it break the class that inherit them.

The other issue is that I no longer have a Pydantic model at that point. They are just Python type descriptors.

As user, I'd like:
- `--use-type-alias` collapse everything but the main model of the schema
- Having another parameter to **opt-in** and collapse the main model too (e.g. `--use-type-alias-for-main-model` or similar)

This way I can switch and use Type Aliases without breaking my current implementation of the generated code. 

BTW the pull request is becoming pretty large. If that's hard/expensive to write down, we can open an issue about that after merging this one :slightly_frowning_face:

docs/root-model-and-type-alias.md
```diff
@@ -0,0 +1,142 @@
...
+class User(BaseModel):
+    id: Optional[UserId] = None
+    status: Optional[Status] = None
+```
+
+## Notes
+
+- RootModel and type aliases do not fully support field-specific metadata (default, alias, etc). See https://docs.pydantic.dev/latest/concepts/types/#named-type-aliases for details.
+- Type aliases do not support some RootModel features (e.g. `model_config`)
+- A RootModel or type alias is also generated for the main schema, allowing you to define a single type alias from a schema file (e.g. `model.json` containing `{"title": "MyString", "type": "string"}`)
```
ilovelinux: Since this section is pretty small, I'd move it after the introduction so the user can acknowledge the limitations of this approach as soon as possible. Otherwise people may read it and then discover that it doesn't work for them

src/datamodel_code_generator/model/template/TypeAliasType.jinja2
butvinm: @ilovelinux I suppose if someone wants to extend generated models, it is also a reasonable use case to extend not only main models, but models from definitions too.

I think what we actually want is fine-grained control on how a model is generated for a specific schema type so we can enable/disable flags only for selected schemas. Maybe in the future we can add support for some comprehensive config file.

I think for now the most reasonable solution is to generate type aliases for types in "definitions" and root models for the main model as you suggested and do not add any additional flags. This way the user can control how to generate a model by moving it to "definitions" or from definitions to a dedicated file.

butvinm: ~Yes, let's fix it in another PR, I am starting to think this PR will never end 😅~

Edit: @ilovelinux I decided to fix it in this PR to not break your (and probably someone else) experience

src/datamodel_code_generator/model/__init__.py
ilovelinux, line 1: Pydantic v1 doesn't support Type Alias. I tested with the example of the issue https://github.com/koxudaxi/datamodel-code-generator/issues/2018 and tried to import the output.

Output of `datamodel-codegen --input input.yaml --input-file-type openapi --use-type-alias`:

```py
#   filename:  input.yaml
#   timestamp: 2025-10-11T18:38:47+00:00

from typing import Optional, Union

from pydantic import BaseModel, constr
from typing_extensions import TypeAliasType

HereBeDragons = TypeAliasType("HereBeDragons", constr(regex=r'here'))


PettyNumber = TypeAliasType("PettyNumber", constr(regex=r'[0-9]{2,8}'))


class PassMe(BaseModel):
    mangled_string: Optional[Union[PettyNumber, HereBeDragons]] = None
```

Which is correct.

But when I try to import it with Pydantic v1.10, I get this:

```py
Python 3.13.7 (main, Aug 15 2025, 12:34:02) [GCC 15.2.1 20250813] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import test
Traceback (most recent call last):
  File "pydantic/validators.py", line 752, in pydantic.validators.find_validators
TypeError: issubclass() arg 1 must be a class

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<python-input-0>", line 1, in <module>
    import test
  File "/tmp/tst/test.py", line 18, in <module>
    class PassMe(BaseModel):
        mangled_string: Optional[Union[PettyNumber, HereBeDragons]] = None
  File "pydantic/main.py", line 197, in pydantic.main.ModelMetaclass.__new__
  File "pydantic/fields.py", line 504, in pydantic.fields.ModelField.infer
  File "pydantic/fields.py", line 434, in pydantic.fields.ModelField.__init__
  File "pydantic/fields.py", line 550, in pydantic.fields.ModelField.prepare
  File "pydantic/fields.py", line 661, in pydantic.fields.ModelField._type_analysis
  File "pydantic/fields.py", line 806, in pydantic.fields.ModelField._create_sub_type
  File "pydantic/fields.py", line 434, in pydantic.fields.ModelField.__init__
  File "pydantic/fields.py", line 555, in pydantic.fields.ModelField.prepare
  File "pydantic/fields.py", line 829, in pydantic.fields.ModelField.populate_validators
  File "pydantic/validators.py", line 761, in find_validators
RuntimeError: error checking inheritance of PettyNumber (type: TypeAliasType)
```

Tried with `--target-python-version 3.12` and I get the same result (as type creates the same object under the hood)

butvinm: We can use `typing.TypeAlias` with pydantic_v1. We will lose type name for schema, but I suppose it is not a big problem (how often someone really use model_json_schema() with pydantic models?). I will mention this constraint in the docs

kapis: > how often someone really use model_json_schema() with pydantic models?

Every single AI framework uses it to generate JSON schemas for OpenAI compatible API. See e.g. [OpenAI SDK](https://github.com/openai/openai-python/blob/e5f93f5daee9f3fc7646833ac235b1693f192a56/src/openai/_compat.py#L177)

ilovelinux: > We can use `typing.TypeAlias` with pydantic_v1. [...]. I will mention this constraint in the docs

Just tested, it works. We can do that then.

Related to https://github.com/pydantic/pydantic/issues/11080

> > how often someone really use model_json_schema() with pydantic models?
> 
> Every single AI framework uses it to generate JSON schemas for OpenAI compatible API. See e.g. [OpenAI SDK](https://github.com/openai/openai-python/blob/e5f93f5daee9f3fc7646833ac235b1693f192a56/src/openai/_compat.py#L177)

I agree with you but, as far as we know, there is no other way to use Type Aliases for Pydantic v1. See https://github.com/pydantic/pydantic/issues/11080

Anyway this feature is opt-in. This means that if you need to use `model_json_schema()` or equivalent, you can just avoid using it.

src/datamodel_code_generator/model/template/TypeAliasType.jinja2
ilovelinux: @butvinm thanks for your effort. I feel like it's okay if the feature is not perfect as it's an opt-in feature. This allows people to not use it if it doesn't fit their needs. Yet, we can improve it in other PRs!

docs/root-model-and-type-alias.md
```diff
@@ -0,0 +1,142 @@
+# Root Models and Type Aliases
+
+When a schema defines a simple type (not an object with properties), `datamodel-code-generator` creates a root model. If you don't want to introduce a new level of attribute access (`.root`) or want to use generated types as plain Python types in non-Pydantic code, you can use the `--use-type-alias` flag to generate type aliases instead of root models.
+
+## Notes
```
ilovelinux:
```suggestion
## :warning: Notes and limitations
```

ilovelinux:
Hi @butvinm, did you have the time to have a look at the Pydantic v1 issue above? I think after fixing that, the PR could be ready to be merged! Other issues/corner-cases will be fixed in other PRs :slightly_smiling_face:

butvinm:
@ilovelinux  I decided not to implement the `--use-type-alias-for-main-model` flag right now and address it in a different PR. Besides that, PR is ready.

butvinm:
Current implementation produces an incorrect type for the model from #2018 
```python
type HereBeDragons = constr(pattern=r'here')
type PettyNumber = constr(pattern=r'[0-9]{2,8}')
```
Call expressions are not allowed in type expressions. We need to generate an annotated str type in such cases. There are probably other similar types which produce the same problem.

I think it should not block the PR and I will fix it in a separate PR after merge.

docs/root-model-and-type-alias.md
```diff
@@ -0,0 +1,252 @@
+# Root Models and Type Aliases
+
+When a schema defines a simple type (not an object with properties), `datamodel-code-generator` creates a root model. If you don't want to introduce a new level of attribute access (`.root`) or want to use generated types as plain Python types in non-Pydantic code, you can use the `--use-type-alias` flag to generate type aliases instead of root models.
+
+## :warning: Notes and limitations
+
```
ilovelinux, line 6:
```suggestion

**This functionality experimental!** Here are a few known issues:
```

docs/index.md
```diff
@@ -387,6 +387,7 @@ Typing customization:
                         Use standard collections for type hinting (list, dict)
   --use-subclass-enum   Define Enum class as subclass with field type when enum has type
                         (int, float, bytes, str)
+  --use-type-alias      Use TypeAlias instead of root models
```
ilovelinux:
```suggestion
  --use-type-alias      Use TypeAlias instead of root models (experimental)
```

README.md
```diff
@@ -395,6 +395,7 @@ Typing customization:
                         Use standard collections for type hinting (list, dict)
   --use-subclass-enum   Define Enum class as subclass with field type when enum has type
                         (int, float, bytes, str)
+  --use-type-alias      Use TypeAlias instead of root models
```
ilovelinux:
```suggestion
  --use-type-alias      Use TypeAlias instead of root models (experimental)
```

src/datamodel_code_generator/arguments.py
```diff
@@ -306,6 +306,12 @@ def start_section(self, heading: str | None) -> None:
     action="store_true",
     default=None,
 )
+typing_options.add_argument(
+    "--use-type-alias",
+    help="Use TypeAlias instead of root models",
```
ilovelinux:
```suggestion
    help="Use TypeAlias instead of root models (experimental)",
```

ilovelinux (changes requested):
Since there are a few things that need to be polished, let's flag this function as experimental so the end user knows that it may not work in their usecase (yet!) 🙂 

After that, I think we're ready to go!

ilovelinux:
@koxudaxi @gaborbernat IMHO this PR is ready to be officially reviewed & merged.

A few tips for the review:
- The feature is **opt-in**, therefore it's not a breaking change.
- Most of the changes are test edits
- As discussed in this PR and written in the documentation, there are a few things that needs to be polished. Because of this the flag has been marked as experimental.
- Known issues be tracked with dedicated issues and addressed in other pull requests.

Thanks for your time :slightly_smiling_face:, and thanks to @butvinm for the PR!

koxudaxi:
@butvinm  @ilovelinux 
Thank you very much for your response.
I finally have time to review this PR.
I noticed the discussion has gotten quite long, but my concern is about the handling of RootModel.
It's very troublesome and I'd like to eliminate it as much as possible, but as you both know, there are edge cases where RootModel becomes necessary. This is why we initially didn't implement RootModel in Pydantic v2, but ended up having to bring it back due to user requests.

I'm a bit concerned that our existing tests may not fully cover the cases where RootModel is absolutely necessary.
Also, as a policy for this project, we make efforts to maintain the same output for the same input as much as possible.
It's likely already being used by many large-scale projects, so even minor changes could generate massive diffs. (Even theoretically safe changes would probably force each project to verify the behavior.)
In other words, ideally we should maintain the same state as before for the same input. (Excluding critical bugs, of course.)

That said, I haven't fully understood the behavior changes in this PR yet, so I'll test it locally to verify the behavior.

butvinm:
@ilovelinux @koxudaxi thanks for your help with PR and review!

ilovelinux:
@koxudaxi 
`TypeAlias` has been deprecated in Python 3.12 in favor of `TypeAliasType`[^1].
I'd keep `TypeAliasType` If there are no issues about that.

[^1]: https://docs.python.org/3/library/typing.html#typing.TypeAlias

koxudaxi:
@ilovelinux 
> Deprecated since version 3.12: [TypeAlias](https://docs.python.org/3/library/typing.html#typing.TypeAlias) is deprecated in favor of the [type](https://docs.python.org/3/reference/simple_stmts.html#type) statement, which creates instances of [TypeAliasType](https://docs.python.org/3/library/typing.html#typing.TypeAliasType) and which natively supports forward references. Note that while [TypeAlias](https://docs.python.org/3/library/typing.html#typing.TypeAlias) and [TypeAliasType](https://docs.python.org/3/library/typing.html#typing.TypeAliasType) serve similar purposes and have similar names, they are distinct and the latter is not the type of the former. Removal of [TypeAlias](https://docs.python.org/3/library/typing.html#typing.TypeAlias) is not currently planned, but users are encouraged to migrate to [type](https://docs.python.org/3/reference/simple_stmts.html#type) statements.

Looking at the documentation, it says `TypeAlias` is deprecated in favor of the type statement (which creates `TypeAliasType` instances), rather than `TypeAliasType` itself. Since the `type` statement isn't available in Python 3.10, I think `TypeAlias` is still the right choice for older Python versions. Does that make sense?

ilovelinux:
@koxudaxi

Facts:
- `TypeAliasType` and `type` are the same thing. The only difference is that `type` is a keyword, which automatically create a `TypeAliasType` instance for that object[^1].
- Ruff recommends to not use `TypeAlias` for Python 3.12+[^2].
- Actually, `TypeAlias` and `TypeAliasType` are not the same thing[^3][^4]
- It's suggested to use `TypeAlias` for Python 3.10-3.11[^5], probably because libraries for that Python version may not expect a `TypeAliasType` object.

It looks like using `TypeAlias` for Python 3.10-3.11 would be better, with an exception for Pydantic v2 which doesn't support it.
However `TypeAliasType` has inherently more informations than `TypeAlias`, which may be useful(?)

[^1]: https://peps.python.org/pep-0695/#runtime-type-alias-class
[^2]: https://docs.astral.sh/ruff/rules/non-pep695-type-alias/
[^3]: https://github.com/python/cpython/issues/114159
[^4]: https://discuss.python.org/t/pep-695-type-aliases-not-suitable-as-replacement-for-typing-typealias/43364
[^5]: https://typing.python.org/en/latest/guides/modernizing.html#type-aliases

koxudaxi:
@ilovelinux 
I used `TypedDict` and this option for a library that supports Python 3.10 to 3.14. In that case, I thought `TypeAlias` would be the safer choice.

ilovelinux:
@koxudaxi 
If we decide to use `TypedDict` for target versions < 3.12, in your case the target version would be Python 3.10 and `TypedDict` would be generated. 

I agree with your proposal to use `TypedDict` until Python 3.12 for all the model types except Pydantic v2

koxudaxi:
@ilovelinux 
Maybe we should have an option to switch between TypeAlias and TypeAliasType? For example, if someone generates dataclasses but wants to use them with Pydantic v2 later. But I'm not sure if this case actually exists, so it might be overkill.

Issue #189, opened 2020-08-01
Support byte of string format


koxudaxi:
This PR supports `byte` of string format.

OpenAPI expects that byte is base64-encoded characters.
https://swagger.io/docs/specification/data-models/data-types/#string

But, I don't know how do we treat encoded/decoded values like base64 in Pydantic.
I opened this issue. However, the problem is not resolved yet.
samuelcolvin/pydantic#692

## Related issues
https://github.com/koxudaxi/datamodel-code-generator/issues/30

kshramt:
https://github.com/pydantic/pydantic/issues/692 was resolved.

PR #2511, opened 2025-10-11
Add support for OpenAPI `byte` in Pydantic v2

ilovelinux:
:warning: Breaking change: byte strings are now automatically decoded by Pydantic v2

Closes https://github.com/koxudaxi/datamodel-code-generator/issues/189


ilovelinux:
@koxudaxi
I think that the point of generating models from schemas is to have models that respects the schema as good as it can. As now, that's not the case for `byte`. Putting the new behavior behind a flag would not introduce a breaking change, but it wouldn't make sense either as this should be the default behavior.

Creating an opt-out configuration option to allow users to not have this behavior may be an idea (as it has been done for https://github.com/koxudaxi/datamodel-code-generator/pull/2512), but I think in this case it would be better to get feedback from the end-users to understand their needs before blindly creating an opt-out flag. We can easily introduce it in the next version if needed :smile: 

A stable release hasn't been released yet, so breaking changes can and should happen to allow the project evolve, introduce needed features/bug-fixes, and one day become stable.

Let me know what do you think about that :slightly_smiling_face:

ilovelinux:
@koxudaxi

> And we should document the breaking change in the release notes page, right?

Yes, this is important.

gaborbernat (approved):

PR #2515, opened 2025-10-11
Add Pydantic v1 to the test matrix

ilovelinux:
We don't run any test about how the code base runs with Pydantic v1, even if it's supposed to be supported.

In fact, the implementations for Pydantic v1 is broken. See #2516.


ilovelinux:
@koxudaxi 
I created #2538, but I couldn't create a PR for this branch because both the branches are on my fork repository.

Luckly, the fixes were very small!

PR #2530, opened 2025-11-03
Support allOf polymorphism with discriminator in OpenAPI

### LINKED ISSUE #2528: OpenAPI polymorphism with discriminator and allOf not working
**Describe the bug**
When discriminator is used with allOf, as described in the OpenAPI doc, no discrimination is applied

**To Reproduce**

Example schema:
```yaml

openapi: 3.1.1
info:
  title: Test
  description: "Test API"
  version: 1.0.0
components:
  schemas:
    Pet:
      type: object
      required:
        - petType
      properties:
        petType:
          type: string
      discriminator:
        propertyName: petType
        mapping:
          cat: '#/components/schemas/Cat'
          dog: '#/components/schemas/Dog'
          lizard: '#/components/schemas/Lizard'
    Cat:
      allOf:
        - $ref: '#/components/schemas/Pet'
        - type: object
          # all other properties specific to a `Cat`
          properties:
            name:
              type: string
    Dog:
      allOf:
        - $ref: '#/components/schemas/Pet'
        - type: object
          # all other properties specific to a `Dog`
          properties:
            bark:
              type: string
    Lizard:
      allOf:
        - $ref: '#/components/schemas/Pet'
        - type: object
          # all other properties specific to a `Lizard`
          properties:
            lovesRocks:
              type: boolean
```

(example schema from directly from [OpenAPI spec](https://spec.openapis.org/oas/v3.1.1.html#examples-1))

Used commandline:
```
$ datamodel-codegen --input openapi.yaml --input-file-type openapi \
    --output-model-type pydantic_v2.BaseModel --output schemas.py \
    --snake-case-field  --use-annotated --use-union-operator --target-python-version '3.11'
```

**Expected behavior**

**Version:**
 - OS: iOS
 - Python version: 3.11
 - datamodel-code-generator version: 35

**Additional context**

See examples in [https://github.com/zdenecek/openapi-discriminator-allof-demo](https://github.com/zdenecek/openapi-discriminator-allof-demo)

zdenecek:
Adds support for use of discriminator with `allOf` as outlined in OpenAPI spec v3.1.1:

[Discriminator Object](https://swagger.io/specification/#discriminator-object).

This MR makes the following example taken directly from OpenAPI spec work as expected:

```yaml
    components:
      schemas:
        Pet:
          type: object
          required:
            - petType
          properties:
            petType:
              type: string
          discriminator:
            propertyName: petType
            mapping:
              dog: Dog
        Cat:
          allOf:
            - $ref: '#/components/schemas/Pet'
            - type: object
              # all other properties specific to a `Cat`
              properties:
                name:
                  type: string
        Dog:
          allOf:
            - $ref: '#/components/schemas/Pet'
            - type: object
              # all other properties specific to a `Dog`
              properties:
                bark:
                  type: string
        Lizard:
          allOf:
            - $ref: '#/components/schemas/Pet'
            - type: object
              # all other properties specific to a `Lizard`
              properties:
                lovesRocks:
                  type: boolean
```

Now if we add a property which references the schema:

```yaml
    PetContainer:
      type: object
      required:
        - pet
      properties:
        pet:
          $ref: "#/components/schemas/Pet"
```

And try to parse it:

```python
from schemas import PetContainer, Cat
import json 

json_data = """
    {
        "pet": {
            "name": "Meow-Meow",
            "petType": "cat"
        }   
    }
    """

container = PetContainer.model_validate_json(json_data)
```

We get what we would expect:

```python
assert isinstance(container.pet, Cat)
assert container.pet.name == "Meow-Meow"
```

This MR adds a test with exactly this schema (from OpenAPI example) while not breaking any other functionallity.

Resolves #2528

### DIFF
--- a/docs/root-model-and-type-alias.md
+++ b/docs/root-model-and-type-alias.md
@@ -10,12 +10,23 @@ When a schema defines a simple type (not an object with properties), `datamodel-
 - A RootModel or type alias is also generated for the main schema, allowing you to define a single type alias from a schema file (e.g. `model.json` containing `{"title": "MyString", "type": "string"}`)
 - Type aliases cannot be combined with `Annotated` for Pydantic v1
 
-## Pydantic v1 vs v2
+## Type Alias Behavior by Output Type and Python Version
 
-The type of type alias generated depends on the Pydantic version:
+The type of type alias generated depends on the output model type and target Python version:
 
-- **Pydantic v2**: Uses `TypeAliasType` (Python 3.9-3.11) or `type` statement (Python 3.12+)
-- **Pydantic v1**: Uses `TypeAlias` because Pydantic v1 cannot handle `TypeAliasType` objects
+| Output Type | Python 3.12+ | Python 3.10-3.11 | Python 3.9 |
+|-------------|--------------|------------------|------------|
+| **Pydantic v2** | `type` statement | `TypeAliasType` (typing_extensions) | `TypeAliasType` (typing_extensions) |
+| **Pydantic v1** | `TypeAlias` | `TypeAlias` | `TypeAlias` (typing_extensions) |
+| **TypedDict** | `type` statement | `TypeAlias` | `TypeAlias` (typing_extensions) |
+| **dataclasses** | `type` statement | `TypeAlias` | `TypeAlias` (typing_extensions) |
+| **msgspec** | `type` statement | `TypeAlias` | `TypeAlias` (typing_extensions) |
+
+**Why the difference?**
+
+- **Pydantic v2** requires `TypeAliasType` because it cannot properly handle `TypeAlias` annotations
+- **Other output types** (TypedDict, dataclasses, msgspec) use `TypeAlias` for better compatibility with libraries that may not expect `TypeAliasType` objects
+- **Python 3.12+** uses the native `type` statement for all output types
 
 ## Example
 
--- a/src/datamodel_code_generator/model/__init__.py
+++ b/src/datamodel_code_generator/model/__init__.py
@@ -55,28 +55,28 @@ def get_data_model_types(
     )
     from .types import DataTypeManager  # noqa: PLC0415
 
-    # Pydantic v1 does not support TypeAliasType type, fallback to TypeAlias
-    if data_model_type == DataModelType.PydanticBaseModel:
-        if target_python_version.has_type_alias:
-            # Python 3.10+: typing.TypeAlias
-            type_alias_class = type_alias.TypeAlias
-            scalar_class = scalar.DataTypeScalar
-            union_class = union.DataTypeUnion
+    # Pydantic v2 requires TypeAliasType; other output types use TypeAlias for better compatibility
+    if data_model_type == DataModelType.PydanticV2BaseModel:
+        if target_python_version.has_type_statement:
+            type_alias_class = type_alias.TypeStatement
+            scalar_class = scalar.DataTypeScalarTypeStatement
+            union_class = union.DataTypeUnionTypeStatement
         else:
-            # Python 3.9: typing_extensions.TypeAlias
-            type_alias_class = type_alias.TypeAliasBackport
-            scalar_class = scalar.DataTypeScalarBackport
-            union_class = union.DataTypeUnionBackport
+            type_alias_class = type_alias.TypeAliasTypeBackport
+            scalar_class = scalar.DataTypeScalarTypeBackport
+            union_class = union.DataTypeUnionTypeBackport
     elif target_python_version.has_type_statement:
-        # Python 3.12+ with Pydantic v2 or other formats: Use type statement
         type_alias_class = type_alias.TypeStatement
         scalar_class = scalar.DataTypeScalarTypeStatement
         union_class = union.DataTypeUnionTypeStatement
+    elif target_python_version.has_type_alias:
+        type_alias_class = type_alias.TypeAlias
+        scalar_class = scalar.DataTypeScalar
+        union_class = union.DataTypeUnion
     else:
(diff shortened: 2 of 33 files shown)


src/datamodel_code_generator/parser/jsonschema.py
ilovelinux, line 1: Since the `discriminator` field is an OpenAPI-only feature, all the code to handle it should go in `openapi.py`

zdenecek: @ilovelinux Thanks, I moved it.

src/datamodel_code_generator/parser/openapi.py
```diff
@@ -318,6 +321,94 @@ def get_data_type(self, obj: JsonSchemaObject) -> DataType:
...
+
+        # Check if this ref has a discriminator but no oneOf/anyOf
+        if ref in self._discriminator_schemas and ref in self._discriminator_subtypes:
+            subtypes = self._discriminator_subtypes[ref]
+            if subtypes:
+                # Create union type with all subtypes
+                data_types = [self.model_resolver.add_ref(subtype_ref) for subtype_ref in subtypes]
+                return self.data_type(data_types=[self.data_type(reference=r) for r in data_types])
+
+        return self.data_type(reference=reference)
```
ilovelinux: The OpenAPI referencing implementation is based on the jsonschema one so we can actually re-use the `jsonschema.py` implementation

```suggestion
        data_type = super().get_ref_data_type(ref)

        # Check if this ref has a discriminator but no oneOf/anyOf
        if ref in self._discriminator_schemas and ref in self._discriminator_subtypes:
            subtypes = self._discriminator_subtypes[ref]
            if subtypes:
                # Create union type with all subtypes
                data_types = [self.model_resolver.add_ref(subtype_ref) for subtype_ref in subtypes]
                return self.data_type(data_types=[self.data_type(reference=r) for r in data_types])

        return data_type
```

src/datamodel_code_generator/parser/openapi.py
```diff
@@ -318,6 +321,94 @@ def get_data_type(self, obj: JsonSchemaObject) -> DataType:
...
+            )
+
+            # If field is a ref to a discriminated schema, add discriminator to extras
+            if field.ref and field.ref in self._discriminator_schemas:
+                discriminator_info = self._discriminator_schemas[field.ref]
+                if discriminator_info:
+                    field_obj.extras["discriminator"] = discriminator_info
+
+            fields.append(field_obj)
+        return fields
```
ilovelinux: I see most of the code is duplicated from `jsonschema.py` but I understand that the OpenAPI logic (L403-407) is in the inner loop so it's harder to not duplicate. I wonder if there's a way to avoid duplicated code...

src/datamodel_code_generator/parser/openapi.py
```diff
@@ -641,3 +734,54 @@ def parse_raw(self) -> None:  # noqa: PLR0912
...
+                self._discriminator_schemas[ref] = {"propertyName": discriminator}
+
+        # Second pass: Find subtypes (schemas that reference discriminated schemas via allOf)
+        for schema_name, schema in schemas.items():
+            if not isinstance(schema, dict):
+                continue
+
+            all_of = schema.get("allOf", [])
+            if not all_of:
+                continue
```
ilovelinux: We can remove that since an empty list already produce the same effect

```suggestion
```

src/datamodel_code_generator/parser/openapi.py
zdenecek: Thanks for the feedback, I see three possible approaches:

1. Keep the code duplicated
    I agree that the downsides to this are probably worse than the other two solutions.
2. Restructure the method into more methods in `jsonschema.py`.
    Do it so that we can then go in and only override the right parts in `openapi.py`
    We could add something like `resolve_field_type`, `post_process_field_object`.
    This is clean DRY-wise, however for the `JsonParser` may seem arbitrary, especially the    
    `post_process_field_object`, making the `JsonParser` code less legible.
3. Call `super().parse_object_fields` and then iterate over the fields again replace the ones in questions with the properly parsed objects. As the code seems free of side effects, this should not introduce any logical issues. The performance impact is negligible and less relevant than downsides of the other options.

I went with n. 3 as the most sensible one. Let me know if this is okay with you or if you think differently.

ilovelinux:
@zdenecek I'll review the code ASAP but as now, it looks pretty good :slightly_smiling_face: 

In the meanwhile, please update your branch. PR have been merged yesterday and your tests may be outdated

src/datamodel_code_generator/parser/openapi.py
ilovelinux, line 1: It looks like there are a lot of dead if branches. E.g. `if not self.raw_obj`.
This is why Codecov is complaining.

Are they really needed? If yes, could you add tests to cover that cases?

src/datamodel_code_generator/parser/openapi.py
```diff
@@ -325,90 +364,51 @@ def get_data_type(self, obj: JsonSchemaObject) -> DataType:
...
-            if subtypes:
-                # Create union type with all subtypes
-                data_types = [self.model_resolver.add_ref(subtype_ref) for subtype_ref in subtypes]
-                return self.data_type(data_types=[self.data_type(reference=r) for r in data_types])
+    def _get_discriminator_union_type(self, ref: str) -> DataType | None:
+        """Create a union type for discriminator subtypes if available."""
+        subtypes = self._discriminator_subtypes.get(ref, [])
+        if not subtypes:
+            return None
+        refs = [self.model_resolver.add_ref(subtype_ref) for subtype_ref in subtypes]
```
ilovelinux: Since it's used in the line below, there's no need to create a list
```suggestion
        refs = map(self.model_resolver.add_ref, subtypes)
```

PR #2550, opened 2025-11-27
Add use_attribute_docstrings Support

### LINKED ISSUE #2049: Add `model_config = ConfigDict(use_attribute_docstrings=True)` when the command option `--use-field-description` is used
The [`use_attribute_docstrings`](https://docs.pydantic.dev/latest/api/config/#pydantic.config.ConfigDict.use_attribute_docstrings) key is available in Pydantic v2.7+ that enables docstrings of attributes to be used for field descriptions. If the value is the default `False`, the examples in the linked documentation would print `None`. It'll be nice for the docstrings to still be accessible via `Model.model_fields["x"].description`, just like if it's defined using `Field(description="Description in Field")`.

voteblake:
Fixes #2049 

In total this change adds support for the upstream [`use_attribute_docstrings`](https://docs.pydantic.dev/latest/api/config/#pydantic.config.ConfigDict.use_attribute_docstrings) for Pydantic 2 models and ~sets it by default when `use_field_description` is configured~ makes it available as a CLI option.

~I've broken this up in to two commits in case enabling it by default when `use_field_description` is not desired.~

Simply having the `use_attribute_docstrings` support would be helpful, as I could then at least configure it by providing extra template data.

I appreciate your consideration.

### DIFF
--- a/README.md
+++ b/README.md
@@ -431,6 +431,8 @@ Field customization:
   --strip-default-none  Strip default None on fields
   --union-mode {smart,left_to_right}
                         Union mode for only pydantic v2 field
+  --use-attribute-docstrings
+                        Set use_attribute_docstrings=True in Pydantic v2 ConfigDict
   --use-default         Use default value even if a field is required
   --use-default-kwarg   Use `default=` instead of a positional argument for Fields that have
                         default values.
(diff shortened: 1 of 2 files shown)


voteblake:
> Thank you for implementing this functionality, it looks very useful! However it may not be needed/wanted by some end users in some situations.
> 
> We can put this feature behind an argument/configuration parameter 🙂

Makes sense, I've pushed a change that adds this. It is a larger change, which I try to avoid for 'drive by' contributions like this, but I believe I've made it the minimal change that accomplishes the goal of making this a CLI option.

PR #2555, opened 2025-11-28
Use time-machine for faster freeze_time

koxudaxi:
## Summary
- Replace freezegun with time-machine library for Python 3.10+
- Fallback to freezegun for Python 3.9 compatibility
- Add session-scoped module preloading fixture


pyproject.toml
ilovelinux, line 81: This is now needed only for `python_version<'3.10'`, isn't it?

tests/conftest.py
```diff
@@ -200,3 +232,13 @@ def _inline_snapshot_file_formats() -> None:
...
+
+
+@pytest.fixture(scope="session", autouse=True)
+def _preload_heavy_modules() -> None:
+    """Pre-import heavy modules once per session to warm up the import cache.
+
+    This reduces per-test overhead when running with pytest-xdist,
+    as each worker only pays the import cost once at session start.
+    """
+    import datamodel_code_generator  # noqa: PLC0415, F401
```
ilovelinux, line 244: Session fixtures in `pytest-xdist` are NOT shared.

See:
- https://github.com/pytest-dev/pytest-xdist/issues/271
- https://github.com/pytest-dev/pytest-xdist/issues/783
- https://github.com/pytest-dev/pytest-xdist/issues/779

Edit: maybe I misread the docstring. If I re-import an already-imported library, it's cached and it's not really re-imported

tests/conftest.py
```diff
@@ -14,6 +15,37 @@
...
+        return datetime.fromisoformat(time_str.replace("Z", "+00:00"))  # pragma: no cover
+
+    def freeze_time(time_to_freeze: str, **kwargs: Any) -> time_machine.travel:  # noqa: ARG001
+        """Freeze time using time-machine (100-200x faster than freezegun)."""
+        dt = _parse_time_string(time_to_freeze)
+        return time_machine.travel(dt, tick=False)
+
+else:
+    from freezegun import freeze_time as freeze_time  # noqa: PLC0414
+
```
ilovelinux, line 48: Can we move this in a autouse fixture?

ilovelinux: Nevermind, there's already an autouse fixture in the main branch

Issue #2565, opened 2025-11-30
incorrect handling of TypeAlias + model_rebuild


raj-open:
## Description ##

The code generator does not parse recursive types correctly.

There are _very_ minor tweaks that would easily correct this.

## To Reproduce ##

Example schema:

```yaml
openapi: 3.0.3
info:
  version: "x.y.z"
  title: 'Some models'
servers:
  - url: "https://google.co.uk"
paths: {}
components:
  schemas:
    Basic:
      nullable: true
      oneOf:
        - type: boolean
        - type: string
        - type: integer

    Recursive:
      oneOf:
        - $ref: "#/components/schemas/Basic"
        - type: array
          items:
            $ref: "#/components/schemas/Recursive"
```

Used command line:

```bash
$ python3 -m datamodel_code_generator \
  --input-file-type openapi \
  --output-model-type pydantic_v2.BaseModel \
  --encoding "UTF-8" \
  --disable-timestamp \
  --use-schema-description \
  --use-standard-collections \
  --use-union-operator \
  --use-default-kwarg \
  --field-constraints \
  --output-datetime-class AwareDatetime \
  --capitalise-enum-members \
  --enum-field-as-literal one \
  --set-default-enum-member \
  --use-subclass-enum \
  --allow-population-by-field-name \
  --strict-nullable \
  --use-double-quotes \
  --use-type-alias \
  --use-field-description \
  --target-python-version 3.14 \
  --input "schema.yaml" \
  --output "models.py"
```

> [!NOTE]
> Replacing 3.14 with 3.12 or 3.13 yields the same results. Under 3.11 there is a slightly different behaviour. In all cases, however, the behaviour is incorrect.

## Observed behaviour ##

The above (for 3.12–14) results in

```py
from __future__ import annotations
type Basic = bool | str | int | None
type Recursive = Basic | list[Recursive]
Recursive.model_rebuild()
```

and for 3.11

```py
from __future__ import annotations
from typing_extensions import TypeAliasType
Basic = TypeAliasType("Basic", bool | str | int | None)
Recursive = TypeAliasType("Recursive", Basic | list[Recursive] | dict[str, Recursive])
Recursive.model_rebuild()
```

## Expected behaviour ##

Under 3.12–14 the correct result would be

```py
from __future__ import annotations
type Basic = bool | str | int | None
type Recursive = Basic | list[Recursive]
```

i.e. no usage the `Recursive.model_rebuild()` command.

[...]

PR #2566, opened 2025-11-30
Bugfix-TypeAlias-Rebuild ---> Main

raj-open:
This is an attempted bugfix in response to #2565.

Briefly, we

1. added a property to indicate that a given model belongs to a type alias
2. in the method `sort_data_models` we guard against adding model paths to `require_update_action_models`, if the model belongs to a type alias

> [!NOTE]
> alternatively to 1., this can also be achieved using `isinstance(..., TypeAliasBase)`)

> [!IMPORTANT]
> in the minor refactor in 2. we consequently prevent paths from being added to the list, if they are already in the list. This was only _partially_ the case in the original code. This needs to be reviewed (e.g. via explorative tests).

### DIFF
--- a/src/datamodel_code_generator/parser/base.py
+++ b/src/datamodel_code_generator/parser/base.py
@@ -138,10 +138,11 @@ def add_model_path_to_list(
     /,
 ) -> list[str]:
     """
-    Auxiliary method which adds model path to list, provided:
+    Auxiliary method which adds model path to list, provided the following hold.
 
     - model is not a type alias
-    - path is not already in the list
+    - path is not already in the list.
+
     """
     if paths is None:
         paths = []


src/datamodel_code_generator/parser/base.py
```diff
@@ -132,6 +132,27 @@ def dump_templates(templates: list[DataModel]) -> str:
...
+    /,
+) -> list[str]:
+    """
+    Auxiliary method which adds model path to list, provided:
+
+    - model is not a type alias
+    - path is not already in the list
+    """
+    if paths is None:
+        paths = list[str]()
```
koxudaxi:
```suggestion
        paths: list[str] = []
```

raj-open: thanks, I am on it

raj-open: @koxudaxi the suggested changes have been pushed incl. coverage of the auxiliary method `add_model_path_to_list`.

koxudaxi:
@raj-open 
Could you add e2e in `tests/main/openapi/test_main_openapi.py` ?
I think the logic should be test with actual openapi schema.

koxudaxi:
@raj-open 

> In which case we just need to add more cases (viz. a recursive type to type_alias.yaml) and update the expected outcome in type_alias.py, right?

Yes, plesae.

pyproject.toml
```diff
@@ -160,8 +160,10 @@ lint.ignore = [
   "C901",   # complex structure
   "COM812", # Conflict with formatter
   "CPY",    # No copyright statements
+  "D200",   # prevent docstrings from being forced to fit one-line
   "D203",   # one-blank-line-before-class (conflicts with D211)
-  "D212",   # multi-line-summary-first-line (conflicts with D213)
+  "D212",   # prevent first newline from being removed in docstrings
+  "D400",   # allow docstrings to contain lines that do not end with a full stop
```
koxudaxi: @raj-open 
I didn't realize the changes. Plesae keep existing docstrings styles.

raj-open: reverted

raj-open: btw. pragma no cover will not skip the test, you will just have to add in the py311 unit test again later (the p312 test for recursive type aliases is still there of course).

koxudaxi (approved):

pyproject.toml
koxudaxi: OK, I will take it over. thank you.

PR #2570, opened 2025-11-30
Fix recursive TypeAlias generation for Python 3.11 (covers mutual alias refs)

koxudaxi:
## Summary
- Add quote-aware rendering for TypeAlias/TypeAliasType so any TypeAlias-to-TypeAlias reference (self or mutual) is emitted as a forward-ref string on Python 3.11, preventing NameError during eager RHS evaluation.
- Update templates to pass collected quote targets; TypeStatement (3.12+) remains unchanged.

## Related Issues
- https://github.com/koxudaxi/datamodel-code-generator/issues/2565#issuecomment-3592411617

## Related PRs
- https://github.com/koxudaxi/datamodel-code-generator/pull/2566

### DIFF
--- a/src/datamodel_code_generator/parser/base.py
+++ b/src/datamodel_code_generator/parser/base.py
@@ -1299,8 +1299,8 @@ class Parser(ABC):
                     model_field.nullable = False
 
     @classmethod
-    def __quote_type_alias_forward_refs(cls, models: list[DataModel]) -> None:
-        """Quote forward references in TypeAlias definitions per PEP 484."""
+    def __update_type_aliases(cls, models: list[DataModel]) -> None:
+        """Update type aliases to properly handle forward references per PEP 484."""
         rendered_aliases: set[str] = set()
 
         for model in models:
@@ -1575,7 +1575,7 @@ class Parser(ABC):
                 if with_import:
                     result += [str(self.imports), str(imports), "\n"]
 
-                self.__quote_type_alias_forward_refs(models)
+                self.__update_type_aliases(models)
                 code = dump_templates(models)
                 result += [code]
 


raj-open:
> [!NOTE]
> The PEP 484 standard <https://peps.python.org/pep-0484/#forward-references> needs to be considered.

An idea would be to alter `DataType`'s `type_hint` property in `src/datamodel_code_generator/types.py` at this point:

```py
            elif self.reference:
                type_ = self.reference.short_name
```

and add a check there to see if up until that point in the transpilation, the type `self.reference.short_name` would have been been rendered. If so, then the above is fine, if not, then it is an instance of a **forward reference** and needs to be quoted.

So somehow, the `.type_hint` method needs to be context aware (NOT of the object currently being handled, but rather of the first object in the recursion stack, since `.type_hint` is a recursion).

I had started to work on a prototype solution,[^commit] but the linter said 'You shall not pass!'. I trust you guys will figure out a good solution.

---

[^commit]: here <https://github.com/raj-open/datamodel-code-generator/commit/87f60500f35f90583572e3430e08424b4eaee8bc> I tried to clean up the logic, and prepare to add context.

koxudaxi:
@raj-open Makes sense. I will fix the expected files. 
Of course, please leave review comments.

koxudaxi:
@raj-open 
I updated the expected values using inline-snapshot.

tests/data/expected/main/openapi/type_alias_forward_ref_multiple.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+class RegularModel(BaseModel):
+    name: Optional[str] = None
+
+
+Third: TypeAlias = str
+
+
+Second: TypeAlias = Union["First", Third, RegularModel]
```
raj-open, line 19: From [PEP 613](https://peps.python.org/pep-0613), which was introduced in 2021, cf. also <https://typing.python.org/en/latest/spec/aliases.html>. So this should be compatible with python 3.10+.

src/datamodel_code_generator/parser/base.py
```diff
@@ -1543,6 +1571,7 @@ class Processed(NamedTuple):
                 if with_import:
                     result += [str(self.imports), str(imports), "\n"]
 
+                self.__quote_type_alias_forward_refs(models)
```
raj-open: Okay, so at the point when this is called, the models are ordered correctly, so that we know which models count as 'already rendered' and which are _forward references_.

> [!IMPORTANT]
> This has to be done here, just before rendering (dump_templates), as 1: we are now guaranteed to know the model order, and 2: we are about to render.

Good.

src/datamodel_code_generator/parser/base.py
```diff
@@ -1294,6 +1294,34 @@ def __set_one_literal_on_default(self, models: list[DataModel]) -> None:
...
+                    if not data_type.reference:
+                        continue
+                    source = data_type.reference.source
+                    if not isinstance(source, TypeAliasBase):
+                        continue
+                    if isinstance(source, TypeStatement):  # pragma: no cover
+                        continue
+                    name = data_type.reference.short_name
+                    if name not in rendered_aliases:
+                        data_type.alias = f'"{name}"'
```
raj-open, line 1325: Since `models` is correctly ordered upon usage, the set `rendered_aliases` keeps track of non-forward references. That makes sense so far.

Sanity check:

- Ultimately we need to set type_hints, which is used in the render and template methods.
- The type_hints are _not yet set_.
- The (nested) type_hints are recursively built upon starting with `.alias` (or `.type`) values.
- By setting/correcting all `.alias` values for all DataType-instances here, it is guaranteed that any latter computation the type hints expressions in templates will be correct.

Good.

src/datamodel_code_generator/parser/base.py
```diff
@@ -1294,6 +1294,34 @@ def __set_one_literal_on_default(self, models: list[DataModel]) -> None:
...
+
+            rendered_aliases.add(model.class_name)
```
raj-open, line 1327: why is this necessary, if the loop has now already been completed and `rendered_aliases` is not a class property?

src/datamodel_code_generator/parser/base.py
```diff
@@ -1294,6 +1294,34 @@ def __set_one_literal_on_default(self, models: list[DataModel]) -> None:
                 if model_field.nullable is not True:  # pragma: no cover
                     model_field.nullable = False
 
+    @classmethod
+    def __quote_type_alias_forward_refs(cls, models: list[DataModel]) -> None:
```
raj-open: maybe this should be renamed to
```py
@classmethod
def update_type_aliases(
    cls,
    modes: list[DataModel],
    /,
) -> None:
    """
    Given a model order updates type aliases in particular to properly handle TypeAliases to conform with PEP 484.
    """
    ...
```

src/datamodel_code_generator/parser/base.py
```diff
@@ -1294,6 +1294,34 @@ def __set_one_literal_on_default(self, models: list[DataModel]) -> None:
...
+        for model in models:
+            if not isinstance(model, TypeAliasBase):
+                continue
+
+            if isinstance(model, TypeStatement):
+                rendered_aliases.add(model.class_name)
+                continue
+
+            for field in model.fields:
+                for data_type in field.data_type.all_data_types:
```
raj-open, line 1315: The only issue I see here - the `data_type`s in `field.data_type.all_data_type` need to have been cloned when added in. Otherwise by changing an alias here, we change the alias of parent instances.

I assume this is the case, as the e2e tests work. Nonetheless,

- [ ] please confirm.

koxudaxi: Confirmed. Each DataType instance is created fresh via `self.data_type(reference=...)` calls (e.g.,　jsonschema.py:753). While the same Reference object may be shared, the DataType instances themselves are　separate. Therefore, modifying data_type.alias only affects that specific DataType instance and does not impact　any parent or other instances.

raj-open: so the "same" field within different type-hint contexts are not the same instance of the field, but clones? Good.

src/datamodel_code_generator/parser/base.py
raj-open: EDIT: I could not determine this properly with the online editor. Just checked out locally and see that the `.add` statement is within the for-loop and at the right place. The above question can be ignored.

raj-open (approved):
desired isolated instances of sub data types confirmed:

`.data_type(reference=...)` is used in the class in `src/datamodel_code_generator/parser/openapi.py` which itself is an extension of `JsonSchemaParser`. In this latter class, the `.data_type(reference=...)` arises from context-methods in `DataTypeManager`.

Issue #2669, opened 2025-12-15
GraphQL parser does not handle renamed objects correctly


siminn-arnorgj:
Title: GraphQL parser does not handle renamed objects correctly

**Describe the bug**
When GraphQL types and inputs get renamed, fields do not use the new name.

**To Reproduce**

Example schema:
```graphql
type lowercasetype {
  foo: Int!
}

type Ref {
  bar: lowercasetype!
}
```

Used commandline:
```
$ cat example.graphql | uv run python -m datamodel_code_generator --input-file-type graphql
```

**Current behavior**

```python
class Ref(BaseModel):
    bar: lowercasetype
    typename__: Optional[Literal['Ref']] = Field('Ref', alias='__typename')


class Lowercasetype(BaseModel):
    foo: Int
    typename__: Optional[Literal['lowercasetype']] = Field(
        'lowercasetype', alias='__typename'
    )
```
**Expected behavior**

```python
class Lowercasetype(BaseModel):
    foo: Int
    typename__: Optional[Literal['lowercasetype']] = Field(
        'lowercasetype', alias='__typename'
    )


class Ref(BaseModel):
    bar: Lowercasetype
    typename__: Optional[Literal['Ref']] = Field('Ref', alias='__typename')
```

**Version:**
 - OS: macOS 26.1
 - Python version: 3.13.9
 - datamodel-code-generator version: from main b458594249e4afca5ddde05c5dd7de48ae39c60b

**Additional context**

This causes an error when trying to use the generated code, as the type `lowercasetype` is not defined.

PR #2670, opened 2025-12-15
Fix GraphQL parser to handle renamed objects correctly

siminn-arnorgj:
## Summary
- Changed _resolve_types in `graphql.py` to use model_resolver.get_class_name rather than the naive type name. This aligns more closely with how the type name is set on the references in the model_resolver and fixes the bug. (Looking more closely at the code, I think graphql.py should be using the reference system on the model_resolver like the other parsers, but I'm not familiar enough with the codebase to be sure that is the right approach).
- In `parse_field`, non-enums now pick up their reference if one exists, so field annotations use the renamed class names instead of using the raw GraphQL name.
- Created tests for the issue and combined with the enum-casing testfile to cover more cases.
- This fix also caused the order of the generated classes to change, so I updated the snapshots accordingly. I'm not familiar enough with the codebase to be sure, but it seems like a step in the right direction to have the dependent classes generated after the classes they depend on.

Fixes: #2669 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved GraphQL type resolution to avoid incorrect enum/Root-type fallbacks and ensure models map to correct types.
  * Ensured forward references in generated models are reliably resolved at runtime.

* **New Features**
  * Added generated models, enums and type aliases to support additional GraphQL casing scenarios.

* **Tests**
  * Updated and added tests validating casing, enum handling, model ordering, and reference resolution.

### DIFF
--- a/src/datamodel_code_generator/parser/graphql.py
+++ b/src/datamodel_code_generator/parser/graphql.py
@@ -484,7 +484,7 @@ class GraphQLParser(Parser):
         obj = graphql.assert_named_type(obj)
         if obj.name in self.references:
             data_type.reference = self.references[obj.name]
-        else:
+        else:  # pragma: no cover
             # Only happens for Query and Mutation root types
             data_type.type = obj.name
 


src/datamodel_code_generator/parser/graphql.py
```diff
@@ -327,7 +327,7 @@ def _resolve_types(self, paths: list[str], schema: graphql.GraphQLSchema) -> Non
                 # TODO: need a special method for each graph type
                 self.references[type_.name] = Reference(
                     path=f"{paths!s}/{resolved_type.value}/{type_.name}",
-                    name=type_.name,
+                    name=self.model_resolver.get_class_name(type_.name, unique=False).name,
```
ilovelinux: Could you elaborate:
- the rationale of this change
- Whether `unique=False` is safe or may impact the code generation

Thanks!

siminn-arnorgj: I just realized this change is not necessary to fix my issue. The rationale was to move the GraphQL parser behaviour more in line with the jsonschema one. I was comparing the field names in the data models after calling `parse_raw` using the GraphQL parser and the jsonschema parser respectively. I noticed that the field names differed for equivalent schemas, and that the name used in the JSON Schema parser came from this line: https://github.com/koxudaxi/datamodel-code-generator/blob/6aa5668e894949b8ebedee9cf62454ea75623a3c/src/datamodel_code_generator/reference.py#L767

The reason this change is not required is that `Parser.__replace_duplicate_name_in_module` is called after parsing, which fixes the names anyway. I will undo the change for now; let me know if this is actually desirable.

The unique flag actually doesn't do anything because there are no references at the model_resolver level.

Issue #2800, opened 2025-12-25
OpenApi extension fields causing recursion on big schemas


ahmetveburak:
**Describe the bug**
Generator cannot process a big schema due to extension fields.
In my case, `x-internalAPI` is causing recursion if there are many components in the schema.

**To Reproduce**

Example schema:
The issue only happens with many components. I cannot share the exact one, but I tried to anonymize it.
https://gist.github.com/ahmetveburak/cd22bcd9666789b86e649e60595d9416

Used commandline:
```
$ datamodel-codegen --input anonymized-schema.json --input-file-type openapi --output model.py
```

**Expected behavior**
Produce schemas successfully.

**Version:**
 - OS: MacOS 26.1
 - Python version: 3.13.9
 - datamodel-code-generator version: 0.48.1

PR #2801, opened 2025-12-25
Exclude OpenAPI/JSON Schema extension fields (x-*)

ahmetveburak:
I'm not sure if this approach meets the design. I had to proceed, and I found this solution.

Fixes: #2800 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved JSON Schema extras handling: vendor extension fields (x-*) are now ignored when detecting schema keywords, preventing unintended schema merging and preserving intended $ref semantics.

* **Tests**
  * Added a test to verify that vendor extension keys do not trigger schema merging, ensuring parsing continues to treat $ref-only schemas correctly.


src/datamodel_code_generator/parser/jsonschema.py
```diff
@@ -463,14 +463,19 @@ def has_ref_with_schema_keywords(self) -> bool:
...
         if not self.ref:
             return False
         other_fields = get_fields_set(self) - {"ref"}
         schema_affecting_fields = other_fields - self.__metadata_only_fields__ - {"extras"}
         if self.extras:
-            schema_affecting_extras = {k for k in self.extras if k not in self.__metadata_only_fields__}
+            # Filter out metadata-only fields AND extension fields (x-* prefix)
+            schema_affecting_extras = {
+                k for k in self.extras if k not in self.__metadata_only_fields__ and not k.startswith("x-")
+            }
```
ilovelinux, line 475: Now `x-*` fields are no longer merged by `$ref`, right?

tests/parser/test_jsonschema.py
```diff
@@ -914,6 +914,28 @@ def test_has_ref_with_schema_keywords_extras_with_metadata_only_keys() -> None:
     assert obj.has_ref_with_schema_keywords is False
 
 
+def test_has_ref_with_schema_keywords_extras_with_extension_keys() -> None:
+    """Test has_ref_with_schema_keywords when extras contains only x-* extension keys.
+
+    OpenAPI/JSON Schema extension fields (x-*) should be treated as metadata
+    and not trigger schema merging, which prevents infinite recursion with
```
ilovelinux, line 922: > infinite recursion

Is recursion for #2800 infinite, or just high enough to hit recursion limits?

ilovelinux (commented):
I'm not sure to understand how this PR affects current behavior and how it fixes the issue. A rationale is missing.

Also see https://github.com/koxudaxi/datamodel-code-generator/issues/2800#issuecomment-3691501141. Maybe this is a workaround for just that case? Maybe RecursionLimit may still be hit but this edit keep recursion under the limit for this particular case?

src/datamodel_code_generator/parser/jsonschema.py
koxudaxi: Yes. After this change, x-* vendor extensions are ignored in has_ref_with_schema_keywords, so they do not trigger `$ref` merge.

tests/parser/test_jsonschema.py
koxudaxi: It is not truly infinite. It recurses many times and finally hits Python recursion limit, so it looks like infinite. This  change avoids that path.

koxudaxi:
@Nekidev Can you provide the input schema? I want to test it.

Nekidev:
@koxudaxi https://github.com/Nekidev/paddle-py/blob/main/openapi.yml

Nekidev:
My temporary fix was to install the library as a library and import it from a script where I had previously set the recursion limit to 5000. It works, but it's not the best solution and neither is it a one-size-fits-all solution. The same repository has a `generate.py` file with it.

koxudaxi:
@Nekidev 
Thank you for reporting the issue. I believe the loop bug had a different cause. I've fixed it.
https://github.com/koxudaxi/datamodel-code-generator/pull/2852

Issue #2887, opened 2026-01-01
use-title-as-name sometimes get ignored


hgl:
**Describe the bug**
For types like in the reproduction, titles are ignored

**To Reproduce**

Example schema:
```js
{
    $schema: "https://json-schema.org/draft/2020-12/schema",
    $defs: {
      Foo: {
        type: "object",
        properties: {
          array: {
            title: "MyName",
            type: "array",
            items: {
              type: "string",
            },
          },
          object: {
            title: "MyObject",
            type: "object",
            additionalProperties: {
              type: "string",
            },
          },
          enum: {
            title: "MyEnum",
            enum: ["foo", "bar"],
          },
          oneOf: {
            title: "MyOneOf",
            oneOf: [
              {
                type: "string",
              },
              {
                type: "number",
              },
            ],
          },
        },
      },
    },
  }
```

Used commandline:
```
[tool.datamodel-codegen]
input-file-type = "jsonschema"
output-model-type = "typing.TypedDict"
target-python-version = "3.13"
use-union-operator = true
skip-root-model = true
use-standard-collections = true
use-frozen-field = true
use-title-as-name = true

```

**Expected behavior**
I'd expect it to create type aliases like

```python
type MyName = list[str]
type MyObject = dict[str, str]
...
```
instead, titles are simply ignored
```python
class Foo(TypedDict):
    array: NotRequired[list[str]]
    object: NotRequired[dict[str, str]]
    enum: NotRequired[Literal['foo', 'bar']]
    oneOf: NotRequired[str | float]
```

**Version:**
 - Python version:
 - datamodel-code-generator version: [e.g. 22]

**Additional context**

PR #2889, opened 2026-01-01
Create type aliases for inline types with title when use-title-as-name is enabled

koxudaxi:
Fixes: #2887

## Summary by CodeRabbit

* **New Features**
  * Creates named aliases for inline types when a title is present, improving generated type names and readability.
  * Better handling of anonymous object schemas with additionalProperties to produce clearer dict typings.
  * Wraps union-typed fields in small named models to preserve titles and adjust default/value parsing semantics.

* **Tests**
  * Added tests validating titled inline type aliasing and the union-wrapper behavior.


hgl:
Thanks again for implementing it so quickly. I tried to build this PR myself, and I couldn't make the unit tests pass. It turned out the `extra` argument used in `model_validate` is only added in `v2.12.5`. Not sure what the solution should be.

koxudaxi:
@hgl 
Thanks for catching this! I've updated the implementation to filter fields manually instead of using `extra="ignore"`, so it now works with all Pydantic v2 versions.

Issue #2919, opened 2026-01-05
Support ClassVar for Pydantic v2


ubaumann:
**Is your feature request related to a problem? Please describe.**
I would like to be able to define Pydantic v2 ClassVars using JSON Schema. 

**Describe the solution you'd like**
After some investigation, I think the cleanest approach is to use a JSON Schema extra property with an `x-` prefix.

Proposed Schema
```
{
    "type": "object",
    "properties": {
        "namespace": {
            "type": "string",
            "x-is_classvar": true,
            "default": "test"
        }
    }
}
```

Is generating a class variable:
```
class Model(BaseModel):
    namespace: ClassVar[str | None] = 'test'
```

**Describe alternatives you've considered**
Another option is to add a new command parameter, but it would be nice to have it inside the schema.

**Additional context**
I am happy to discuss and create a PR

PR #2920, opened 2026-01-05
Support ClassVar for Pydantic v2

ubaumann:
Fixes: #2919 

While analysing a suitable approach, I was close to an implementation, so I already created the PR. I am happy to discuss the PR and make adjustments. 
How should this "magic" behaviour be documented if this is a possible approach?
Thanks for your effort.

## Summary by CodeRabbit

* **New Features**
  * Models now preserve ClassVar (class-level) attributes with correct ClassVar type hints and default values for Pydantic v1 and v2 outputs; these remain class attributes, not model fields.

* **Tests**
  * Added tests and expected fixtures covering ClassVar handling across multiple Pydantic v2 generation scenarios (including annotated and set forms).

### DIFF
--- a/src/datamodel_code_generator/model/pydantic_v2/base_model.py
+++ b/src/datamodel_code_generator/model/pydantic_v2/base_model.py
@@ -14,6 +14,7 @@ from pydantic import Field
 
 from datamodel_code_generator.imports import IMPORT_ANY, Import
 from datamodel_code_generator.model.base import ALL_MODEL, UNDEFINED, BaseClassDataType, DataModelFieldBase
+from datamodel_code_generator.model.imports import IMPORT_CLASSVAR
 from datamodel_code_generator.model.pydantic.base_model import (
     BaseModelBase,
 )
@@ -31,7 +32,6 @@ from datamodel_code_generator.model.pydantic_v2.imports import (
     IMPORT_VALIDATION_INFO,
     IMPORT_VALIDATOR_FUNCTION_WRAP_HANDLER,
 )
-from datamodel_code_generator.model.imports import IMPORT_CLASSVAR
 from datamodel_code_generator.reference import ModelResolver
 from datamodel_code_generator.types import chain_as_tuple
 from datamodel_code_generator.util import field_validator, model_validate, model_validator


src/datamodel_code_generator/model/pydantic_v2/base_model.py
```diff
@@ -187,11 +188,21 @@ def _has_discriminator_in_data_type(self) -> bool:
         """Check if any nested DataType has a discriminator."""
         return any(dt.discriminator for dt in self.data_type.all_data_types)
 
+    @property
+    def class_var_type_hint(self) -> str:
+        return f"ClassVar[{self.type_hint}]"
+
+    @property
+    def is_class_var(self) -> bool:
+        return self.extras.get("x-is_classvar") is True
```
koxudaxi: Should be `x-is-classver` ? But, I don't have the strong opinion.
What do you think?

src/datamodel_code_generator/model/template/pydantic_v2/BaseModel.jinja2
```diff
@@ -16,7 +16,9 @@ class {{ class_name }}({{ base_class }}):{% if comment is defined %}  # {{ comme
 {%- endfilter %}
 {%- endif %}
 {%- for field in fields %}
-    {%- if not field.annotated and field.field %}
+    {%- if field.is_class_var and field.represented_default %}
+    {{ field.name }}: {{ field.class_var_type_hint }} = {{ field.represented_default }}
+    {%- elif not field.annotated and field.field %}
```
koxudaxi: I don't want to update the template for new feature. Because, users will need to update the custom template.
Can we apply the type hint in the python level?

src/datamodel_code_generator/model/pydantic_v2/base_model.py
ubaumann: I don't have strong opinions either. I just let me be inspired by https://github.com/koxudaxi/datamodel-code-generator/blob/58e73ed8740d589623647008bb1fe230bac1aeda/src/datamodel_code_generator/model/msgspec.py#L86

I think the dash makes it look better.

src/datamodel_code_generator/model/template/pydantic_v2/BaseModel.jinja2
ubaumann: Make sense. I will try to find a way to do it in the code

src/datamodel_code_generator/model/pydantic_v2/base_model.py
koxudaxi: OK, I understand. I like dash .But, We will be able to accpet `-` and `_` in the future.

ubaumann: I changed it to `-`

src/datamodel_code_generator/model/template/pydantic_v2/BaseModel.jinja2
ubaumann: I implemented it in the code and reverted the template

koxudaxi (approved):

Issue #2785, opened 2025-12-24
Add ryaml as optional YAML parser dependency


koxudaxi:
## Summary

  [ryaml](https://github.com/emmatyping/ryaml) is a Rust-based YAML parser that offers better performance than PyYAML. Development has resumed with a comprehensive roadmap (see [ryaml#10](https://github.com/emmatyping/ryaml/issues/10)). Once the next version is released, we should consider adding it as an optional dependency.

  ## Background

  - ryaml is a Rust-based YAML parser with potential performance benefits
  - Development was on hiatus but has now resumed
  - Planned improvements include:
    - Python 3.13+ and free-threading support
    - Migration to actively maintained YAML parser (Saphyr)
    - Broader platform support via cibuildwheel
    - YAML spec conformance testing

  ## Proposal

  Once a stable release is available:

  1. Add `ryaml` as an optional dependency in `pyproject.toml`
     ```toml
     [project.optional-dependencies]
     ryaml = ["ryaml>=X.Y.Z"]

  2. Implement runtime detection to use ryaml when available, falling back to PyYAML
  3. Benchmark performance improvements for large schema files

  Blocked By

  - Next stable release of ryaml with the roadmap improvements

  References

  - ryaml repository: https://github.com/emmatyping/ryaml
  - Development roadmap: https://github.com/emmatyping/ryaml/issues/10

PR #2928, opened 2026-01-05
Migrate from pyright to ty type checker

koxudaxi:
## Summary by CodeRabbit

* **Chores**
  * Updated development tooling and internal configuration for project maintenance.


ilovelinux:
_Premise: I have no experience with `ty` yet._

I partially agree with @nbro10, `ty` is a bleeding edge tool and it isn't going to compete with more stable and vastly (functionally) better tools. But it's a type checker, it's not going to impact that much the development of the project (except code quality in the long term).

I appreciate using the new bleeding edge tools, as it has been done replacing `pre-commit` with `prek` and adding support for a new YAML parsing library proposed in #2785. However it may be a good idea to try bleeding-edge tools like type checkers & linters in tandem with more stable and widely supported tools, at least a certain point of the development (e.g. for non-draft PRs, only in CI)

@koxudaxi how's your experience been so far?

Issue #2964, opened 2026-01-13
use_union_operator=True does not work with specific field names


matejscerba:
**Describe the bug**

When using flag `use_union_operator=True` and specific field names (e.g. `float`), that are not required, Pydantic typing fails and the generated model can not be used. Setting `use_union_operator` to `False` works as expected.

**To Reproduce**

Example schema:
```yaml
openapi: 3.1.0
paths: {}
components:
  schemas:
    ReservedProperty:
      type: object
      properties:
        float:
          type: number
          format: float
```

Used commandline:
```
$ datamodel-codegen --input <path to the file above>
```

The command above generates the following content:
```python
from __future__ import annotations

from pydantic import BaseModel


class ReservedProperty(BaseModel):
    float: float | None = None
```

When trying to instantiate the class (`a = ReservedProperty()`), I get the following error:
```
TypeError: unsupported operand type(s) for |: 'NoneType' and 'NoneType'
```

It appears that field name `float` clashes with `float` type.

**Expected behavior**

I expect the generated file to be usable within Python without any errors.

**Version:**
 - OS: macOS Tahoe 26.1
 - Python version: 3.13.5
 - datamodel-code-generator version: 0.52.2

**Additional context**

PR #2968, opened 2026-01-19
Fix use_union_operator with Python builtin type field names

koxudaxi:
Fixes: #2964

## Summary by CodeRabbit

* **Bug Fixes**
  * Field names that conflict with Python built-ins are now sanitized: internal attribute names receive a trailing underscore while preserving original names as public aliases. This applies to primitives, container and complex builtin names, and respects target-Python-version differences.

* **Tests**
  * Added and expanded tests covering builtin-name collisions (primitives, containers, snake_case, x-python-type cases) and target-Python-version behaviors.


tests/main/openapi/test_main_openapi.py
```diff
@@ -4712,6 +4712,20 @@ def test_main_enum_builtin_conflict(output_file: Path) -> None:
...
+    with freeze_time(TIMESTAMP):
+        run_main_and_assert(
+            input_path=OPEN_API_DATA_PATH / "builtin_type_field_names.yaml",
+            output_path=output_file,
+            input_file_type="openapi",
+            assert_func=assert_file_content,
+            expected_file="builtin_type_field_names.py",
+            extra_args=["--output-model-type", "pydantic_v2.BaseModel"],
+        )
+
```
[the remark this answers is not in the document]
koxudaxi: This is not needed because `pydantic_v2.BaseModel` uses `from __future__ import annotations` by default, which enables the union operator syntax automatically.

src/datamodel_code_generator/parser/base.py
```diff
@@ -111,6 +111,35 @@ def __ge__(self, value: Any, /) -> bool: ...  # noqa: D105
...
+    "reversed",
+    "set",
+    "slice",
+    "staticmethod",
+    "str",
+    "super",
+    "tuple",
+    "type",
+    "zip",
+})
```
ilovelinux: What about:
```py
import builtins

_BUILTIN_NAMES: frozenset[str] = frozenset(dir(builtins))
```

It also includes builtin classes and it's automatically updated[^1].

[^1]: https://docs.python.org/3/library/builtins.html

src/datamodel_code_generator/parser/base.py
```diff
@@ -111,6 +111,35 @@ def __ge__(self, value: Any, /) -> bool: ...  # noqa: D105
 ModelDeps: TypeAlias = dict[ModelName, set[ModelName]]
 OrderIndex: TypeAlias = dict[ModelName, int]
 
+_BUILTIN_NAMES: frozenset[str] = frozenset({
```
ilovelinux: Static declaration doesn't honor `--target-python-version`. E.g. some keyword may/may not be included in `builtins`.
However I don't have any good idea to keep it simple and make it honor the targeted python version.

tests/data/expected/main/openapi/builtin_type_field_names.py
ilovelinux, line 1: Shouldn't the leading underscore be applied to all the builtins keywords[^1]?

[^1]: https://peps.python.org/pep-0008/#descriptive-naming-styles

src/datamodel_code_generator/parser/base.py
koxudaxi: Thanks, I switched `_BUILTIN_NAMES` to derive from `builtins.__dict__` so the set updates automatically and includes builtin classes/functions as suggested. This is now in commit a5ce41ba.

tests/data/expected/main/openapi/builtin_type_field_names.py
koxudaxi: I kept the underscore rename collision-based rather than applying it to every builtin identifier, to avoid broad alias churn for common fields like `type` where there is no type-shadowing conflict. This PR focuses on builtin *type* collision safety and now also covers the snake_case (`Float` -> `float_`, alias preserved) regression.

src/datamodel_code_generator/parser/base.py
koxudaxi: Good point. This still cannot perfectly mirror every `--target-python-version`, but the rename itself remains collision-based (`dt.type`/container checks), so behavior is bounded to actual type-name collisions. I kept this PR scoped to that and added a regression test for snake_case alias preservation.

src/datamodel_code_generator/parser/base.py
```diff
@@ -111,6 +112,30 @@
...
+        case "set":
+            is_container_match = data_type.is_set
+        case "frozenset":
+            is_container_match = data_type.is_frozen_set
+        case "tuple":
+            is_container_match = data_type.is_tuple
+        case _:
+            is_container_match = False
+
+    return is_container_match
```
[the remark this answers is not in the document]
koxudaxi: Fixed in f956271b. `is_container_match` is now initialized before the `match` statement, so all control-flow paths have a defined value before return.

Issue #2976, opened 2026-02-02
Enum default=0 is not converted to enum member with --set-default-enum-member


kkinugasa:
**Describe the bug**
When using `--set-default-enum-member`, an `IntEnum` field default value `0` is not converted to the corresponding enum member (e.g. `Status.integer_0`). The generated model keeps the numeric default (`0`) instead of `Status.integer_0`.

**To Reproduce**

Example schema:
```json
{
  "$schema": "https://json-schema.org/draft/2020-12/schema",
  "type": "object",
  "properties": {
    "status": {
      "$ref": "#/$defs/Status",
      "default": 0
    }
  },
  "$defs": {
    "Status": {
      "type": "integer",
      "enum": [0, 1]
    }
  }
}
```

Used commandline:
```console
$ datamodel-codegen --input test.json --set-default-enum-member
```

Output:
```python
from __future__ import annotations

from enum import IntEnum

from pydantic import BaseModel


class Status(IntEnum):
    integer_0 = 0
    integer_1 = 1


class Model(BaseModel):
    status: Status | None = 0
```

**Expected behavior**
The default value should be converted to the corresponding enum member when `--set-default-enum-member` is enabled.

```python
class Model(BaseModel):
    status: Status | None = Status.integer_0
```

**Version:**
 - OS: macOS
 - Python version: 3.13.11
 - datamodel-code-generator version: 0.53.1

**Additional context**
Confirmed locally: the issue is fixed for this case by applying the following changes.

  In https://github.com/koxudaxi/datamodel-code-generator/blob/365419ec30ebba7cbad5c1c8c7f66a08cfefacbd/src/datamodel_code_generator/parser/base.py#L2039:
  ```diff
  - if not model_field.default:
  + if model_field.default is None:
  ```

  In https://github.com/koxudaxi/datamodel-code-generator/blob/365419ec30ebba7cbad5c1c8c7f66a08cfefacbd/src/datamodel_code_generator/model/enum.py#L106:
  ```diff
  - field_default = str(field.default or "").strip("'\"")
  + field_default = str("" if field.default is None else field.default).strip("'\"")
  ```

ilovelinux:
I see you already have a fix proposal. PRs are welcome!

PR #2977, opened 2026-02-02
Fix handling of falsy default values for enums in set-default-enum-member option

kkinugasa:
Fixes: #2976 

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected enum member matching so falsy defaults (e.g., 0, empty string, False) are honored rather than treated as absent.

* **Tests**
  * Added test coverage validating default enum member selection for falsy default values in JSON Schema generation.

* **Samples**
  * Added expected output illustrating enums and a model using falsy default enum members.

### DIFF
--- a/tests/data/expected/main/jsonschema/falsy_default_enum_member.py
+++ b/tests/data/expected/main/jsonschema/falsy_default_enum_member.py
@@ -19,6 +19,12 @@ class Name(Enum):
     non_empty = 'non-empty'
 
 
+class Active(Enum):
+    boolean_True = True
+    boolean_False = False
+
+
 class Model(BaseModel):
     status: Status | None = Status.integer_0
     name: Name | None = Name.field_
+    active: Active | None = Active.boolean_False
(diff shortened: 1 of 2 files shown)


tests/main/jsonschema/test_main_jsonschema.py
```diff
@@ -3299,6 +3299,19 @@ def test_main_jsonschema_modular_default_enum_member(output_dir: Path) -> None:
         )
 
 
+def test_main_jsonschema_falsy_default_enum_number(output_file: Path) -> None:
+    """Test enum member mapping for falsy default values."""
+    with freeze_time(TIMESTAMP):
```
ilovelinux: I wonder why we used this in the other enum test

src/datamodel_code_generator/model/enum.py
```diff
@@ -103,7 +103,7 @@ def find_member(self, value: Any) -> Member | None:
 
         for field in self.fields:
             # Remove surrounding quotes from field default value
-            field_default = str(field.default or "").strip("'\"")
+            field_default = str("" if field.default is None else field.default).strip("'\"")
```
ilovelinux: I'd simplify this a little bit
```suggestion
            field_default = "" if field.default is None else str(field.default).strip("'\"")
```

ilovelinux (changes requested):

ilovelinux (dismissed):
Good job! LGTM!

PR #3026, opened 2026-03-05
Remove pydantic v1 output support

koxudaxi:


docs/dynamic-model-generation.md
```diff
@@ -292,7 +292,7 @@ assert all(r is results[0] for r in results)
 | Use case | Code generation, file output | Runtime validation |
 | Caching | No | Yes (configurable) |
 | Thread-safe | Yes | Yes |
-| Pydantic v1 | Yes | No |
+| Pydantic v2 | Yes | Yes |
```
ilovelinux, line 295: This may be removed since it's now supported on Pydantic v2 only

docs/what_is_the_difference_between_v1_and_v2.md
ilovelinux, line 1: Name of this file is misleading since we no longer support Pydantic v1

Issue #3035, opened 2026-03-08
--use-type-alias: inline object unions validate defaults through the first branch only


keyz:
**Describe the bug**
`datamodel-code-generator==0.54.1` generates an incorrect `default_factory` for inline object unions: it always validates through the first branch.

**To Reproduce** 

Example schema:
```json
{
  "$defs": {
    "A": { "properties": { "type": { "const": "a" } } },
    "B": { "properties": { "type": { "const": "b" } } }
  },
  "properties": {
    "x": {
      "default": { "type": "b" },
      "anyOf": [
        { "$ref": "#/$defs/A" },
        { "$ref": "#/$defs/B" }
      ]
    }
  }
}
```

Used commandline:
```
$ datamodel-codegen \
    --input schema.json \
    --input-file-type jsonschema \
    --output out.py \
    --output-model-type pydantic_v2.BaseModel \
    --use-type-alias \
    --target-python-version 3.12
```

Observed output:

```python
class Model(BaseModel):
    x: A | B | None = Field(default_factory=lambda: A.model_validate({"type": "b"}))
```

Observed behavior:

- mypy passes
- Model() fails at runtime because {"type": "b"} is validated as A

Expected:

- validate the default against the union and produce B, not hardcode branch 1

clawdney:
Hi! I'm interested in helping with this. Could you share more context about the use case?

PR #3040, opened 2026-03-08
fix #3034 and #3035 by introducing `ValidatedDefault`

### LINKED ISSUE #3034: --use-type-alias emits raw dict defaults for $ref object unions
**Describe the bug**
`datamodel-code-generator==0.54.1` emits an invalid default for a `$ref` to an object union when `--use-type-alias` is enabled

**To Reproduce**

Example schema:
```json
{
  "$defs": {
    "A": { "properties": { "type": { "const": "a" } } },
    "B": { "properties": { "type": { "const": "b" } } },
    "U": {
      "anyOf": [
        { "$ref": "#/$defs/A" },
        { "$ref": "#/$defs/B" }
      ]
    }
  },
  "properties": {
    "x": {
      "default": { "type": "b" },
      "$ref": "#/$defs/U"
    }
  }
}
```

Used commandline:
```
$ datamodel-codegen \
    --input schema.json \
    --input-file-type jsonschema \
    --output out.py \
    --output-model-type pydantic_v2.BaseModel \
    --use-type-alias \
    --target-python-version 3.12
```

Observed output:
```python
type U = A | B

class Model(BaseModel):
    x: U | None = {"type": "b"}
```
Observed behavior:

- mypy reports the default as incompatible
- Model().x is a raw dict at runtime, not A | B

Expected:

- emit a validated model default, not a raw dict literal

keyz:
this is an attempt to fix #3034 and #3035 via introducing `ValidatedDefault` for structured defaults (that must be validated through the declared field type)

before this change, some union and type-alias defaults were emitted either as raw dict/list literals or validated through the wrong union branch

overall I think the fixture coverage is good, but there may be a cleaner fix than `ValidatedDefault`. please feel free to edit this PR or open a separate PR if you see a better solution!

---

as extra context: here are the CLI flags I use. I use the library to translate [codex app-server json schema](https://developers.openai.com/codex/app-server/#message-schema) to pydantic models:

```
--input-file-type jsonschema
--output-model-type pydantic_v2.BaseModel
--target-pydantic-version 2.11
--target-python-version 3.12
--use-default-kwarg
--snake-case-field
--use-union-operator
--use-standard-collections
--field-constraints
--use-annotated
--reuse-model
--collapse-reuse-models
--enum-field-as-literal all
--use-title-as-name
--allow-population-by-field-name
--use-type-alias
```

## Summary by CodeRabbit

* **New Features**
  * Generated models can embed runtime-validated default values for complex types (unions, type aliases, lists, dicts), ensuring defaults conform to declared field types.
  * Introduces an opt-in capability flag to enable validated defaults; Pydantic v2 support reflects this capability.

* **Tests**
  * Large test suite added to cover many type-alias and default-value scenarios across JSON Schema patterns and configuration combinations.

### DIFF
--- a/tests/parser/test_base.py
+++ b/tests/parser/test_base.py
@@ -676,6 +676,20 @@ def test_contains_model_reference_traverses_nested_data_types() -> None:
     assert _contains_model_reference(list_of_models) is True
 
 
+def test_contains_model_reference_traverses_dict_key() -> None:
+    """Test _contains_model_reference walks dict_key data types."""
+    model_reference = Reference(path="KeyModel", original_name="KeyModel", name="KeyModel")
+    BaseModel(fields=[], reference=model_reference)
+
+    dict_with_model_key = DataType(
+        is_dict=True,
+        dict_key=DataType(reference=model_reference),
+        use_union_operator=True,
+    )
+
+    assert _contains_model_reference(dict_with_model_key) is True
+
+
 def test_get_validated_default_type_name_strips_optional_none() -> None:
     """Test _get_validated_default_type_name removes None from optional types."""
     optional_string = DataType(type="str", is_optional=True, use_union_operator=True)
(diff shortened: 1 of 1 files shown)


koxudaxi (changes requested):
Please check the comment

src/datamodel_code_generator/parser/base.py
```diff
@@ -399,6 +400,92 @@ def iter_models_field_data_types(
...
+def _needs_validated_default(default: Any, data_type: DataType) -> bool:
+    if not isinstance(default, (dict, list)):
+        return False
+
+    resolved = _unwrap_type_alias(data_type)
+    if not _contains_model_reference(resolved):
+        return False
+
+    if resolved.is_union:
+        non_none_branches = tuple(branch for branch in data_type.data_types if branch.type_hint != NONE)
```
koxudaxi: Use `resolved.data_types` here instead of `data_type.data_types`.
If data_type is a reference to a type alias that expands to a union, `data_type.data_types` can be empty, while `resolved.data_types` contains the actual union branches.

koxudaxi: @keyz 
I pushed a small correctness fix for the `resolved.data_types` issue to help unblock it.
The remaining items are the validated-default type timing and the model-level capability cleanup.

keyz:
@codex review

PR #3054, opened 2026-03-16
Fix required field default rendering and --use-default nullable types

### LINKED ISSUE #3048: Default values for required fields are sometimes rendered and sometimes not
Default values for required fields are rendered for lists of objects, nested objects and dictionaries but not for plain fields like strings, ints and enums.

Example schema:
```json

{
  "$schema": "http://json-schema.org/draft-07/schema",
  "title": "Order",
  "type": "object",
  "definitions": {
    "Status": {
      "type": "string",
      "enum": ["pending", "processing", "shipped", "delivered"]
    },
    "Priority": {
      "type": "integer",
      "enum": [1, 2, 3]
    },
    "Address": {
      "type": "object",
      "properties": {
        "street": {"type": "string"},
        "city": {"type": "string"}
      },
      "required": ["street", "city"]
    }
  },
  "properties": {
    "order_id": {"type": "string"},
    "status": {
      "$ref": "#/definitions/Status",
      "default": "pending"
    },
    "priority": {
      "allOf": [{"$ref": "#/definitions/Priority"}],
      "default": 2
    },
    "quantity": {
      "type": "integer",
      "default": 1
    },
    "note": {
      "type": "string",
      "default": "your note here"
    },
    "tags": {
      "type": "array",
      "items": {"type": "string"},
      "default": ["new"]
    },
    "metadata": {
      "type": "object",
      "additionalProperties": {"type": "string"},
      "default": {"source": "web"}
    },
    "shipping_address": {
      "allOf": [{"$ref": "#/definitions/Address"}],
      "default": {"street": "123 Main St", "city": "Springfield"}
    },
    "all_addresses": {
      "type": "array",
      "items": {"$ref": "#/definitions/Address"},
      "default": [{"street": "1 First St", "city": "Shelbyville"}]
    },
    "addresses": {
      "type": "object",
      "additionalProperties": {"$ref": "#/definitions/Address"},
      "default": {"home": {"street": "10 Oak Ave", "city": "Ogdenville"}}
    }
  },
  "required": ["order_id", "status", "priority", "quantity", "note", "tags", "metadata", "shipping_address", "all_addresses", "addresses"]
}


```

Used commandline:
```
datamodel-codegen --input schema.json --input-file-type jsonschema --output-model-type pydantic_v2.BaseModel --output model.py
```

Generated models:

```
#   filename:  pydantic_v2_model_defaults.json
#   timestamp: 2019-07-26T00:00:00+00:00

from enum import Enum, IntEnum

class Status(Enum):
    pending = 'pending'
    processing = 'processing'
    shipped = 'shipped'
    delivered = 'delivered'


class Priority(IntEnum):
    integer_1 = 1
    integer_2 = 2
    integer_3 = 3


class Address(BaseModel):
    street: str
    city: str


class Order(BaseModel):
    order_id: str
    status: Status
    priority: Priority
    quantity: int
    note: str
    tags: list[str]
    metadata: dict[str, str]
    shipping_address: Address = Field(
        default_factory=lambda: Address.model_validate(
            {'street': '123 Main St', 'city': 'Springfield'}
        )
    )
    all_addresses: list[Address] = Field(
        default_factory=lambda: [
            Address.model_validate(v)
            for v in [{'street': '1 First St', 'city': 'Shelbyville'}]
        ]
    )
    addresses: dict[str, Address] = Field(
        default_factory=lambda: {
            k: Address.model_validate(v)
            for k, v in {'home': {'street': '10 Oak Ave', 'city': 'Ogdenville'}}.items()
        }
    )

```

**Expected behavior**
In the generated models some of the fields have the defaults, some not. I would expect all the fields to have defaults.

**Version:**
 - OS: N/A
 - Python version: 3.13
 - datamodel-code-generator version: 0.55.0

**Additional context**

It's probably related to https://github.com/koxudaxi/datamodel-code-generator/issues/437

There is a flag `--use-default` but it works by making fields non-required so this behaviour is not desired.

butvinm:
Fixes #3048

Two changes:

**1. Consistent required field defaults (fixes the reported inconsistency)**

`__set_validate_default_on_fields` now skips required fields, so model-ref fields no longer get defaults while scalar fields don't. All required fields behave the same — no defaults rendered without `--use-default`.

**2. `--use-default` no longer makes fields nullable (breaking change)**

Previously `--use-default` turned `status: str | None = 'active'` — making required fields optional and nullable. Now it produces `status: str = 'active'` — default is rendered but the type stays non-nullable. I consider the previous behavior a bug: the schema says the type is `string`, not `string | null`, and `--use-default` shouldn't change that.

## Before (bug)

Using the schema from #3048, without `--use-default`:
```python
class Order(BaseModel):
    order_id: str
    status: Status
    priority: Priority
    quantity: int
    note: str
    tags: list[str]
    metadata: dict[str, str]
    shipping_address: Address = Field(
        {'street': '123 Main St', 'city': 'Springfield'}, validate_default=True
    )
    all_addresses: list[Address] = Field(
        [{'street': '1 First St', 'city': 'Shelbyville'}], validate_default=True
    )
    addresses: dict[str, Address] = Field(
        {'home': {'street': '10 Oak Ave', 'city': 'Ogdenville'}}, validate_default=True
    )
```
Scalar required fields (`status`, `quantity`, etc.) have defaults dropped, but model-ref required fields (`shipping_address`, etc.) render defaults — inconsistent.

## After (fix)

Without `--use-default` — consistent, all required fields have no defaults:
```python
class Order(BaseModel):
    order_id: str
    status: Status
    priority: Priority
    quantity: int
    note: str
    tags: list[str]
    metadata: dict[str, str]
    shipping_address: Address
    all_addresses: list[Address]
    addresses: dict[str, Address]
```

With `--use-default` — all defaults rendered, types stay non-nullable:
```python
class Order(BaseModel):
    order_id: str
    status: Status = 'pending'
    priority: Priority = 2
    quantity: int = 1
    note: str = 'your note here'
    tags: list[str] = ['new']
    metadata: dict[str, str] = {'source': 'web'}
    shipping_address: Address = Field(
        {'street': '123 Main St', 'city': 'Springfield'}, validate_default=True
    )
    all_addresses: list[Address] = Field(
        [{'street': '1 First St', 'city': 'Shelbyville'}], validate_default=True
    )
    addresses: dict[str, Address] = Field(
        {'home': {'street': '10 Oak Ave', 'city': 'Ogdenville'}}, validate_default=True
    )
```

## Summary by CodeRabbit

* **New Features**
  * Option to apply default values to fields while keeping them required and non-nullable.

* **Bug Fixes**
  * Preserve default/default_factory for required fields when the option is enabled.
  * Avoid emitting placeholder-only representations for required fields when defaults are used.

* **Tests**
  * Added E2E tests and updated expected fixtures covering required-with-default behavior across JSON Schema, OpenAPI, and GraphQL.


koxudaxi:
@ilovelinux If you have time, could you share your thoughts on this PR, the related issues, and the approach? I feel like the PR needs to be split since it addresses two separate problems.

ilovelinux (commented):
Hi @koxudaxi, @butvinm. Sorry for the late answer. Last week has been tough.

> [!NOTE]
> I splitted this comment in multiple paragraph. I though I wouldn't have much to say, but I was wrong. I hope they are easier to read.

### About the changes

> [!NOTE]
> @koxudaxi 
> > If you have time, could you share your thoughts on this PR, the related issues, and the approach?
>
> I admit reviewing changes hasn't been easy. I feel like we should reduce code complexity to increase code maintainability. However it kinda makes sense to me. I'll try to do a more in-depth review after this comment to double-check the logic.

I appreciate @butvinm work and agree with the fixes introduced, because:
- The first change allow a coherent usage of default values.
- The second change fix `None` type hint for required properties with default value defined (which can't be `None` because they are required).

IMHO both changes make sense and reflect the expected user experience.

I also found an useful description in the OpenAPI docs[^1] (emphasis mine):
> The default value to use for substitution, which SHALL be **sent** if an alternate value is not supplied.

E.g. We have a Python client and a Python server which both uses `datamodel-code-generator`. The server shouldn't populate ANY property with default values, instead the client should use default values and send them explicitly.

More generally, this adhere to the DRY principle: one side populate the data, the other side validates the data. Double-validation and, more important, double-population of default values, is error-prone.

As now, it isn't possible to adhere to the DRY principle using models generated by `datamodel-code-generator`. This PR make default values coherent avoiding double-population, and removes wrong `None` type hint improving data validation, allowing to apply DRY principle.

### About the multiple-responsibilities

@koxudaxi 

> It looks like the PR is trying to fix multiple problems at once, so I want to check the related issues for each of them before reviewing.

I agree the two changes solve two different problems. However:
1. I'm not sure if they are also independent of each other.
    - If they are interdependent, splitting would make things more complex since one of them would depend on the other.
3. They are both breaking changes about the same topic and I feel like they should be released together to reduce the end-user impact.

Despite single-responsibility PRs are cleaner and easier to review, I think we could avoid splitting this PR because the two changes are small and share the same surface impact (= default values).

### About the breaking changes

I don't know how could a migration period for the users looks like for a change like this. I am not sure it would be useful because there's no way to avoid the breaking change. E.g.: changing the default formatter is a breaking change, but the migration period allow the users to acknowledge that and pin the old ones in the configuration file; this, instead, changes the way default values are rendered. There's no way to avoid that for the user and there's nothing the user can do to mitigate the breaking change. This looks more like an "incompatible version migration".

What if, instead, we communicate to the user what did change as soon as the user run the new version? A warning about breaking changes would:
- Reduce the cognitive effort to keep track of the breaking changes.
- Increases the chances the user understand and acknowledge what and why that happened.
- Encourage the user to open an issue if something is broken because of the breaking change (e.g. a corner case that is no longer handled).

In this case, I think the users would appreciate this PR since it's a bugfix and I don't think many people were relying on either the wrong generated type hints or the default values only for model-ref fields. I think, instead, many other people may be having some hard time because of this.

---

I'm not approving this PR because I'm taking some time to think about the impact, what @koxudaxi said, and all I have written. In the meanwhile, I'd like to hear what do you all think! 🙂 

Also, a hot pizza is waiting for me! 🍕 Can't ignore Italian priorities 😆 

[^1]: https://swagger.io/specification/#server-variable-object

butvinm:
Hi! Sorry for disappearing — it was a tough month. I agree with limiting the scope of this PR to the two fixes, and moving the broader default-handling discussion to a separate issue.

Merged latest main, caught and fixed a few implementation bugs. PR is ready to merge whenever you'd like.

I'll open the follow-up issue for the broader default-handling discussion soon.

PR #3083, opened 2026-04-14
Prefer CLI input over pyproject url

koxudaxi:
Fixes: #3076

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved input source precedence to ensure CLI-provided arguments correctly override configuration file settings, preventing conflicts from multiple input sources.


vdusek:
Hi @koxudaxi, I think this should apply to all fields, not just `("input", "url", "input_model")`.

In general, users should be able to configure every field either through a configuration file or via CLI arguments, with CLI arguments taking precedence.

koxudaxi:
@vdusek 
Thanks. I agree that CLI options should generally take precedence over config file values.

In this case, though, that already happens for normal fields. The special handling here is only for `input`, `url`, and `input_model` because they are mutually exclusive. Without clearing the other two, values from `pyproject.toml` can remain and conflict with the CLI input source.

vdusek:
@koxudaxi Perfect, thanks for the clarification.

PR #3086, opened 2026-04-14
Add flag --use-default-keep-required-non-nullable

GuillaumeSmaha:
Hi!

The aim is to provide a way to keep the field non nullable/optional when --use-default is used

```yaml
    Pet:
      required:
        - id
        - name
      properties:
        id:
          type: integer
          format: int64
          default: 1
        name:
          type: string
        tag:
          type: string
```
wil become:
```python
class Pet(BaseModel):
    id: int = 1
    name: str
    tag: str | None = None
```
instead of (current implementation):
```python
class Pet(BaseModel):
    id: int | None = 1
    name: str
    tag: str | None = None
```

## Summary by CodeRabbit

* **New Features**
  * Added `--use-default-keep-required-non-nullable` CLI option: when used with `--use-default`, required fields remain non-nullable/non-optional instead of becoming nullable.

* **Documentation**
  * Updated CLI reference and quick-reference with the new option and expanded examples showing OpenAPI input plus Pydantic v2, dataclass, msgspec, and JSON Schema output tabs.

* **Tests**
  * Added CLI and generation tests and new expected model fixtures covering the new behavior.


src/datamodel_code_generator/parser/jsonschema.py
```diff
@@ -694,10 +694,8 @@ def __init__(
         self._circular_ref_cache: dict[str, bool] = {}
 
         if self.data_model_field_type.can_have_extra_keys:
-            self.get_field_extra_key: Callable[[str], str] = (
-                lambda key: self.model_resolver.get_valid_field_name_and_alias(
-                    key, model_type=self.field_name_model_type
-                )[0]
+            self.get_field_extra_key: Callable[[str], str] = lambda key: (
```
GuillaumeSmaha, line 697: `uv run ruff format` did this update

GuillaumeSmaha:
Ahah I just realize that `--strict-nullable` is almost the same. When I checked it, I was getting few different result.
I am going to put it back to draft in the mean time

PR #3178, opened 2026-05-23
Add builtin formatter warning

koxudaxi:
## Summary by CodeRabbit

* **New Features**
  * Added a dependency-free "builtin" formatter and a CLI/config option to control its line-length and behavior.

* **Documentation**
  * Updated docs, CLI help, pyproject guidance, site navigation, and migration notes describing formatter choices and builtin behavior.

* **Tests**
  * Added extensive unit/integration tests and optional parity checks comparing builtin vs Black+isort outputs.

* **Chores / CI**
  * Added a new test environment and updated CI/test matrix and tooling configs.

* **Deprecations**
  * Clarified that default external formatters will become opt-in in a future release.

[Review Change Stack](https://app.coderabbit.ai/change-stack/koxudaxi/datamodel-code-generator/pull/3178?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


ilovelinux:
This looks related to #3058. Great idea adding a simple built-in formatter!

koxudaxi:
@ilovelinux 
Yes, exactly. I considered several options, but for this issue, switching to ruff doesn't really solve the problem.

I'm honestly not sure how far a built-in formatter can go in handling general formatting like black does, but I'd like to try.

Ideally, if we can get output equivalent to black or ruff, it would be a breaking change, but the barrier to removing black/isort from dependencies would be much lower.

ilovelinux:
@koxudaxi 

I'm thinking out loud here about the builtin formatter. I'm a bit hesitant about how far we should take it, mainly for a few reasons:
- **Maintenance & Bugs**: Building out a complex formatter introduces new code to maintain and edge cases to fix.
- **Separation of Duties**: Formatting should ideally be handled by dedicated tools, or a separate library. Generating unformatted code (unless an external formatter is installed) makes sense as formatting may be considered out of scope. 

If we do move forward with a builtin formatter, I'd highly recommend keeping it as simple as possible, perhaps just wrapping exceptionally long generated lines for basic readability. If users want strict formatting, they can install `black`, `ruff`, etc.

I suspect the goal here is to replace `black`/`isort` without creating a massive diff. While I understand that motivation, I'm not sure the added complexity is worth it since there's no functional impact on the codebase.

> [!NOTE]
> The message has been rewritten by AI to improve clarity, then reviewed by me.
>
><details><summary>Original message</summary>
><p>
> I'm thinking out loud...
>
> I don't think the builtin formatter should go very far. I'm still unsure if that's worth the effort since it is going to introduce code to maintain and some potential bugs. Formatting the output code should be handled by a separate component because of separation of duties. Generating an unformatted code unless a formatter is installed may still be an option.
> But if we're going forward with the builtin formatter, I think it should just improve readability for those very long lines that might be generated. I'd keep it simple. If the user wants a better formatter, they just need to install them (e.g. black, isort, ruff, etc...).
>
>However, I think you are trying to implement the builtin formatter in this specific way because you'd like to replace black & isort without generating a big diff. I'm not sure it's worth since it's not going to have a functional impact on the codebase.
> </p>
> </details>

koxudaxi:
@ilovelinux 
Thanks for pointing this out.

I changed the implementation so the internal formatter is not used as a pre-formatter for black/isort/ruff. It is now only an alternative formatter when no external formatter is selected.

If `builtin` is specified together with an external formatter, the external formatter wins and `builtin` is ignored with a warning. This avoids adding extra processing cost for users who choose ruff/black/isort, especially in the future if ruff is selected explicitly.

The goal of the internal formatter is not full compatibility with black/isort/ruff for arbitrary Python code. It only tries to cover the code patterns generated by datamodel-code-generator, so we can reduce differences from the current black+isort output while keeping the implementation small.

I also added tests to check compatibility against generated expected outputs and kept the current default behavior unchanged for existing users.

koxudaxi:
@ilovelinux 
I don't want to force another review, but if you have time, could you check whether this is within an acceptable range?
I think it should be fine if we release it with an experimental label.

src/datamodel_code_generator/format.py
```diff
@@ -191,13 +196,1275 @@ def black_find_project_root(sources: Sequence[Path]) -> Path:
...
+
+def _alias_sort_key(alias: str) -> tuple[int, str]:
+    name = _alias_imported_name(alias)
+    if name.isupper():
+        category = 0
+    elif name[:1].isupper():
+        category = 1
+    else:
+        category = 2
+    return category, name.lower()
```
ilovelinux, line 328: Can we use `IntEnum` here to improve readability?

src/datamodel_code_generator/format.py
```diff
@@ -191,13 +196,1275 @@ def black_find_project_root(sources: Sequence[Path]) -> Path:
...
+    if module == "__future__":
+        return 0
+    if level:
+        return 4
+    top_level_module = module.split(".", 1)[0]
+    if top_level_module in sys.stdlib_module_names:
+        return 1
+    if top_level_module in known_first_party:
+        return 3
+    return 2
```
ilovelinux: Same here

src/datamodel_code_generator/format.py
```diff
@@ -191,13 +196,1275 @@ def black_find_project_root(sources: Sequence[Path]) -> Path:
...
+    string_normalization: bool = False,
+) -> str:
+    """Apply dependency-free formatting for generated Python code."""
+    lines = [line.rstrip() for line in code.splitlines()]
+    code = "\n".join(lines).strip("\n")
+    if not code:
+        return ""
+
+    try:
+        tree = ast.parse(code)
```
ilovelinux: We should use `feature_version` respecting `--target-python` value.

See: https://docs.python.org/3/library/ast.html#ast.parse

ilovelinux (commented):
@koxudaxi 

I wrote some minor comments below.

I think some tests that format all the test outputs we have may help us to check that:
- The formatter doesn't crash.
- The output is syntactically correct.

It would be great to add the last check for generation outputs too.

Maybe using https://docs.python.org/3/library/concurrent.interpreters.html to import the generated module.

PR #3210, opened 2026-05-28
Require Pydantic 2.6

### LINKED ISSUE #3202: Minimum Pydantic Version
**Describe the bug**
Running `datamodel-codegen` (with no arguments) with a sufficiently old (<=2.5.3, but still considered supported) Pydantic version throws `pydantic.errors.PydanticUndefinedAnnotation: name 'Optional' is not defined` from running `model_rebuild` on https://github.com/koxudaxi/datamodel-code-generator/blob/0.58.0/src/datamodel_code_generator/model/base.py#L1009-L1011.

**To Reproduce**
- Install Pydantic 2.5.3 or older.
- Run `datamodel-codegen`.

Used commandline:
```
$ datamodel-codegen
```

**Expected behavior**
It should show the usage options (and then work when given real input) instead of crashing.

**Version:**
 - OS: Ubuntu 24.04
 - Python version: 3.12
 - datamodel-code-generator version: 0.58.0

**Additional context**
I was using an older version of datamodel-code-generator (0.25.3) and an older version of Pydantic. I decided to upgrade datamodel-code-generator to resolve another problem, but I wasn't planning on upgrading Pydantic for now. It didn't work. Simply upgrading to Pydantic 2.6.0 or newer works, though. I didn't identify the earliest version of datamodel-code-generator that required the Pydantic upgrade. The minimum supported version should be updated and documented.

koxudaxi:
Fixes: #3202

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -35,7 +35,7 @@ dependencies = [
   "inflect>=4.1,<8",
   "isort>=4.3.21,<9",
   "jinja2>=2.10.1,<4",
-  "pydantic>=2,<3; python_version<'3.14'",
+  "pydantic>=2.6,<3; python_version<'3.14'",
   "pydantic>=2.12,<3; python_version>='3.14'",
   "pyyaml>=6.0.1",
   "tomli>=2.2.1,<3; python_version<'3.11'",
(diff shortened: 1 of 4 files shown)


koxudaxi:
@coderabbitai review

Issue #3212, opened 2026-05-28
--type-overrides affecteds more fields then intended


supervirus:
**Describe the bug**

When using scoped type-overrides to change the type of a single field other fields (of the same original type) are affected, too.

The aim is simply to override the type of a single field in a model or class from `int` (a Unix timestamp) to `datetime.datetime`.
However, when using type overrides, all fields of source type `integer` in the class are suddenly changed to `datetime`.

**To Reproduce**

Example schema:
```yaml

openapi: '3.0.3'

components:
  schemas:
    BasketInfo:
      type: object
      properties:
        name:
          type: string
          description: Basket name
          example: basket22
        requests_count:
          type: integer
          description: Current number of collected HTTP requests held by basket
          example: 62
        requests_total_count:
          type: integer
          description: Total number of all HTTP requests passed through this basket
          example: 315
        last_request_date:
          type: integer
          format: int64
          description: |
            Date and time of last request processed through this basket in Unix time ms.
            format (number of milliseconds elapsed since January 1, 1970 UTC).

            If no requests were collected by this basket `0` is returned.
          example: 1550106301288

```

Used commandline:
```
$ datamodel-codegen
```

Used pyproject.toml
```toml
[project]
name = "bugreport"
version = "0.1.0"
description = "Add your description here"
requires-python = ">=3.14"
dependencies = [
    "pydantic>=2.13.4,<3",
]

[dependency-groups]
dev = [
    "datamodel-code-generator[http,ruff,validation]",
]


[tool.ruff.format]
# Use `\n` line endings for all files
line-ending = "lf"

[tool.ruff.lint]
extend-select = ["I"]

[tool.ruff.lint.isort]
# Use a single line after each import block.
lines-after-imports = 2

[tool.datamodel-codegen]
target-python-version = "3.14"
use-double-quotes = true
use-generic-container-types = true
use-standard-collections = true
use-union-operator = true
field-constraints = true
use-annotated = true
formatters = ["ruff-check", "ruff-format"]
type-overrides = {"BasketInfo.last_request_date" = "datetime.datetime"}
input = "test-schema.yaml"
# Full model: url = "https://raw.githubusercontent.com/darklynx/request-baskets/refs/heads/master/doc/rbaskets-openapi.yaml"
input-file-type = "openapi"
output-model-type = "pydantic_v2.BaseModel"
output = "models.py"
```

**Expected behavior**
[...]

PR #3213, opened 2026-05-28
Fix scoped type-overrides changing more fields then intended

supervirus:
Fixes scoped type-overrides (#3212) by creating a deepcopy of field.data_type before changing it.
Without the deepcopy() the whole data_type is modified, which affects all other fields using this data_type, too.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented type override changes from unintentionally affecting shared type instances elsewhere, ensuring overrides are applied safely without side effects.

[Review Change Stack](https://app.coderabbit.ai/change-stack/koxudaxi/datamodel-code-generator/pull/3213?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/datamodel_code_generator/parser/base.py
+++ b/src/datamodel_code_generator/parser/base.py
@@ -15,6 +15,7 @@ import sys
 from abc import ABC, abstractmethod
 from collections import Counter, OrderedDict, defaultdict
 from collections.abc import Callable, Hashable, Mapping, Sequence
+from copy import deepcopy
 from itertools import groupby
 from pathlib import Path
 from typing import (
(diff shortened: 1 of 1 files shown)


supervirus:
Hi,
I am unsure if I am using the generation store as intended:

```python
        data_type = deepcopy(field.data_type)
        data_type.import_ = override_import
        data_type.alias = override_import.import_
        self.generation_store.replace_field_type(field, data_type)

```

And if there calls are even still necessary / correct after replacing the field type:
```python
        self.generation_store.detach_data_type_ref(field.data_type)
        self.generation_store.set_nested_data_types(field.data_type, [])
```

Please feel free to make changes as necessary or let me know what I need to change.

Thanks.

Issue #2414, opened 2025-06-08
Use validators for models that have "one of" requirements.


jkugler:
**Is your feature request related to a problem? Please describe.**
When a model has a "oneOf" requirement, Datamodel creates multiple models, with each field required, and then creates a "root" model that merges all those models. This makes attribute access painful.

**Describe the solution you'd like**
Instead of creating multiple models, use Pydantic validators to make sure on one of the fields has a value, and raise an error if not the case.

**Describe alternatives you've considered**
I have modified the template for Root models to add `__getattr__` and `__setattr__` methods. But this would not be needed if it didn't create multiple models.

jbbqqf:
@jkugler — bumping the workaround @butvinm suggested above (2026-03-07): `--collapse-root-models` and `--use-type-alias` are still in `main` and target exactly this `RootModel`-wrapping pain point.

Verified at `src/datamodel_code_generator/arguments.py`:

```
model_options.add_argument(
    "--collapse-root-models",
    action="store_true",
    ...
    help="Models generated with a root-type field will be merged into the models using that root-type model",
)
typing_options.add_argument(
    "--use-type-alias",
    help="Use TypeAlias instead of root models (experimental)",
    action="store_true",
    ...
)
```

There is also `--collapse-root-models-name-strategy` (`child`/`parent`) for naming preferences. If either flag resolves the `.root` indirection without needing the validator-based redesign, this could be closed as `wontfix → use existing flags`. If you've tested and they don't quite fit (e.g. you still need the runtime validation aspect of the original ask), please leave a follow-up so the maintainers know the gap is real.

---
*Disclosure: Drive-by triage from a GCP/Python data engineer running an OSS-triage pass; not affiliated with the project. AI-assisted comment, evidence verified by hand.*

jkugler:
I'll try to give it a go next week (currently OOO).

koxudaxi:
Closing as addressed by #3228, released in 0.66.1.

The new opt-in `--generate-schema-validators` / `--schema-validator-type pydantic-v2` path emits Pydantic v2 runtime validators for required-only `oneOf`/`anyOf` groups, covering the validator-based approach requested here.

For the separate RootModel attribute indirection case, `--collapse-root-models` or `--use-type-alias` remain available. If there is still a specific `oneOf` shape not covered, please open a new issue with a minimal schema and expected output.

PR #3228, opened 2026-06-01
Add schema validators

### LINKED ISSUE #1312: the generated class  missing attributes/features when jsonschema is defined with allof + patternProperties
**Describe the bug**
generated class  missing attributes/features when jsonschema is defined with allof + patternproperties

**To Reproduce**

Example schema: (in yaml)
```yaml
definitions:
  First:
    properties:
      first:
        type: string
    required:
      - first
  Second:
    properties:
      second:
        type: string
    required:
      - second

  Third:
    patternProperties:
       ^([a-zA-Z][a-zA-Z0-9_]+)$:
          $ref: '#/definitions/Second'

  Target:
    allOf:
      - $ref: "#/definitions/First"
      - $ref: "#/definitions/Third"

  AnotherTarget:
    allOf:
      - $ref: "#/definitions/First"
      - patternProperties:
           ^([a-zA-Z][a-zA-Z0-9_]+)$:
              $ref: '#/definitions/Second'

```

Used commandline:
```
$ datamodel-codegen  --input-file-type jsonschema      --input input.yaml
```
generated result:
```python
:
class First(BaseModel):
    first: str
class Second(BaseModel):
    second: str
class Third(BaseModel):
    __root__: Dict[constr(regex=r'^([a-zA-Z][a-zA-Z0-9_]+)$'), Second]
class Target(First):
    pass
class AnotherTarget(First):
    pass
```

**Expected behavior**
generated class 'Target' should have attributes/features for patternProperties(#/definitions/Third) but missing.
AnotherTaget  class also expected as the same as Target.

**Version:**
 - OS:  debian11
 - Python version: 3.11
 - datamodel-code-generator version:  latest master branch (8855a13)

**Additional context**
none.

### LINKED ISSUE #1875: Consolidate Data Models for "oneOf"["required.. JSON Schema models

One of the reasons why I want to use pydantic instead of json-schema (where the source of truth for data models I'm using lie) is because if much easier error codes and information when data doesn't validate.  However, when translating a json schema that specifies that one of or any of a set of requirements must be met, it creates a bunch of separate data models and puts them together in a single data model that is a union.  This results in error codes that are....not very helpful.  It creates a laundry list of all the possible sub models that aren't being met. 

```json
...
"oneOf": [
   {"required":["option_1"]},
   {"required":["option_2"]},
   {"required":["option_3"]},
...
```

```python

class ParentModel1(BaseModel):
    other_info:....
    option_1: Option1
    option_2: Optional[Option2] = None
    option_3: Optional[Option3] = None

class ParentModel2(BaseModel):
    other_info:....
    option_1: Optional[Option1] = None
    option_2: Option2
    option_3: Optional[Option3] = None

class ParentModel3(BaseModel):
    other_info:....
    option_1: Optional[Option1] = None
    option_2: Optional[Option2] = None
    option_3: Option3

class ParentModel(
    RootModel[
        Union[
            ParentModel1,
            ParentModel2,
            ParentModel3, #note that in my case I have a few dozen of these...
        ]
    ]
):
    root: Annotated[
        Union[
            ParentModel1,
            ParentModel2,
            ParentModel3,
        ],
        Field(title='Parent Schema'),
    ]

```

Translation of a "oneOf" statement (and similar) to a pydantic @rootmodel validator.

After spending a few hours trying to accurately post-process the resulting data models, I currently edit the resulting data models by hand, which isn't traceable or reproducible in a data pipeline.

I also did a search to see if other issues like this had been opened and couldn't immediately find any - which was surprising to me so maybe I missed something?

**Additional context**

### LINKED ISSUE #2386: Support if/then/else from JSON Schema Core
JSON Schema Core has the keywords `if`, `then` and `else for conditional schema validation. 

Currently datamodel-codegen seems to ignore these fields within a JSON Schema.

Ideally when generating a pydantic model from a schema that contains these keywords I expect a generated model that enforces the requirements imposed by these keywords.

https://datatracker.ietf.org/doc/html/draft-handrews-json-schema-02#section-9.2.2.1

https://json-schema.org/understanding-json-schema/reference/conditionals#ifthenelse

koxudaxi:
Closes: #1312
Closes: #2386
Closes: #1875

Related: #2414

## Summary by CodeRabbit

* **New Features**
  * Added --generate-schema-validators CLI option to emit Pydantic v2 runtime validators for JSON Schema (patternProperties, conditional required properties, oneOf/anyOf groups).

* **Documentation**
  * Updated CLI reference, quick reference, index, and full docs with the new option, usage examples, and generated-output samples; adjusted template customization index/count.

* **Bug Fixes**
  * CLI now rejects the new option when used with non-Pydantic-v2 output types with a clear error.

* **Tests**
  * Added unit and end-to-end tests covering generated validators and CLI behavior.


tests/data/expected/main/jsonschema/schema_validators.py
```diff
@@ -0,0 +1,244 @@
...
+class RequiredOnlyGuard1(BaseModel):
+    flag: str | None = None
+
+
+class RequiredOnlyGuard(RootModel[RequiredOnlyGuard1]):
+    root: RequiredOnlyGuard1
+
+
+class Model(RootModel[PatternTarget]):
+    root: PatternTarget
```
koxudaxi, line 244: @ilovelinux 
If you get a chance, I'd love your opinion on this runtime validation code generation. I went through a lot of options, but this is the only solid idea I could come up with for something reusable and general-purpose.

gaborbernat (approved):
LGTM

gaborbernat (dismissed):
LGTM

tests/data/expected/main/jsonschema/schema_validators.py
ilovelinux: @koxudaxi
I'm very sorry for the late answer.
I once needed to fully validate the data against the jsonschema and I did that creating another class they inherits the genrared one and overriding the __init__ to call the  jsonschema library's validation function before deserialization. It was a few lines long and it was very reliable.
I'm worried that adding responsibilities like this instead of delegate them to external dedicated libraries may be a burden to maintain in the long term. Once we merge this, users may ask to have the same thing implemented for the other formats too...
Instead, I'd add a flag to validate data against the jsonschema using a dedicated, highly performante and fully tested library that the user will put among the dependencies. This already happens when you generate Pydantic models or using some parameters like pendulum: generated code has external dependencies.
Implementation may or may not look like the one I described above. It also has the advantage of being portable among different output model types. This, instead, looks like a feature that is going to be available only for Pydantic v2.
I'll write down an example of the workaround I used.

koxudaxi: @ilovelinux 
Thanks, I think your point is valid.

One nuance is that the current implementation is not only pure JSON Schema validation.
For patternProperties / additionalProperties, it also validates matched extra values
through Pydantic's TypeAdapter, so those values can be coerced/deserialized into the
generated Python/Pydantic types. A dedicated JSON Schema validator would be a better
fit for full spec-compliant validation, but by itself it would only validate the raw
input and would not provide that Pydantic-specific conversion for dynamic extra keys.

That said, I agree that maintaining these runtime validation semantics can increase
the long-term maintenance burden. Still, I would like users to be able to choose the
runtime validation behavior explicitly over time.

So instead of replacing this approach with an external JSON Schema validator, maybe
we can treat them as different validator backends/options: one backend integrated
with Pydantic v2 that can validate and coerce dynamic extra values, and another
backend that delegates validation to a dedicated JSON Schema library for users who
prefer full JSON Schema validation through an external dependency.

What do you think?

gaborbernat (approved):
LGTM

gaborbernat (approved):
LGTM

gaborbernat (approved):
LGTM

gaborbernat (dismissed):
LGTM

koxudaxi:
Thanks for the repeated reviews, and sorry for the delay.
I haven't had much time to work on this over the last few days. As mentioned above, I'd like to update the CLI first, so I'll work on that later today.

tests/data/expected/main/jsonschema/schema_validators.py
koxudaxi: Addressed in 73835eb4. This PR now adds an explicit `--schema-validator-type` enum selector with the current `pydantic-v2` backend, keeps `--generate-schema-validators` as the compatibility path, and covers the CLI/public API path with e2e tests using the existing assert helpers and external fixtures. I left external validator backend integration out of this PR so it can be added separately if needed. CI is green.

PR #3237, opened 2026-06-03
Allow leading underscore class names

### LINKED ISSUE #3235: Support Names That Start With Underscores

I am using `generate_dynamic_models()` with user-provided JSON Schemas and I need to generate an internal/root model with a reserved name that cannot realistically collide with user-defined model names.

I tried using:

```python
GenerateConfig(
    class_name="__ParsedModel",
    ...
)
```

However, names starting with underscores are sanitized/changed by the generator. For example, `__ParsedModel` does not remain `__ParsedModel`.

This makes it difficult to safely distinguish the internally generated/root parsed model from models that may come from user-provided schemas. In dynamic model generation, this is especially important because the generated models are returned in a dictionary by class name.

The problem is that I cannot reliably reserve a special internal model name such as `__ParsedModel`.

I would like `class_name` to support explicitly provided class names starting with underscores, especially in dynamic model generation.

For example:

```python
GenerateConfig(
    class_name="__ParsedModel",
    ...
)
```

should generate and return a model whose `__name__` is exactly:

```python
"__ParsedModel"
```

Ideally, `generate_dynamic_models()` should also be able to return models with names starting with `_` when they were explicitly requested by the user through `class_name`.

Possible approaches:

1. Preserve explicitly provided `class_name` values exactly, even if they start with `_`.
2. Add a config option such as:

```python
preserve_class_name=True
```

or:

```python
allow_leading_underscore_class_name=True
```

3. Add a config option specifically for dynamic generation to include underscore-prefixed generated classes in the returned dictionary.

The important part is that explicitly requested class names should not be silently changed when the user intentionally provides them.

I considered generating the model with a random public names, like

```python
random_internal_name = f"ParsedMode{secrets.token_hex(3)}"
GeneratedModel = models[random_internal_name]
```

but that call to `secrets.token_hex` wastes compute and the PascalCase formatting can pollute the name, making it non-deterministic, from the user's POV

**Additional context**

This is mainly useful when `datamodel-code-generator` is used as a runtime/dynamic schema compiler for user-provided JSON Schemas.

In that case, users may define arbitrary schema titles and model names, so the application needs a safe internal name for the root parsed model. A double-underscore name like `__ParsedModel` is useful because it is unlikely to be accidentally provided by users and clearly marks the model as internal.

I think it would be a cool feature to have!

**PS**

Sorry for raising the issue if that feature is already somehow implemented, but I couldn't find anywhere in the docs how to accomplish that

koxudaxi:
Closes: #3235


koxudaxi:
@coderabbitai review

koxudaxi:
@coderabbitai The lock-scope nit is intentionally left unchanged in this PR. `generate_dynamic_models()` already serialized generation and module execution before this change; this PR only threads the opt-in class-name preservation flag and keeps existing concurrency semantics intact. Changing the single-module path to double-check locking would be a separate performance/concurrency refactor, especially because the multi-module path mutates `sys.modules`.

PR #3329, opened 2026-06-11
Deduplicate parser base helpers

koxudaxi:


koxudaxi:
Keeping stacklevel=1 intentionally: this helper preserves the duplicated pre-refactor warning behavior, and the parser/base refactor requires the exact warning message and stacklevel.

PR #3353, opened 2026-06-12
Fix `constr()` for string fields carrying minItems/maxItems

DarkaMaul:
A pydantic v2 string field whose schema uses `minItems`/`maxItems` generated `constr(min_items=..., max_items=...)`. `constr` has no `min_items`/`max_items` parameters , so building the model raises `TypeError: constr() got an unexpected keyword argument 'min_items'`

Add a regression test and fixture.

## Release Notes

* **Bug Fixes**
  * Fixed JSON Schema constraint translation for Pydantic v2. String fields with `minItems`/`maxItems` constraints are now correctly mapped to appropriate length validation rules instead of incompatible item-count constraints.


koxudaxi (changes requested):
Thanks for the fix. The implementation direction looks good, but CI is failing because this fixture is now picked up by the schema-derived payload validation tests.

`minItems` / `maxItems` are valid JSON Schema keywords, but they are inapplicable to `type: string`. So payloads like `{"code": ""}` are valid against the source JSON Schema, while the generated Pydantic model now rejects them via `constr(min_length=2, max_length=5)`.

Could you please update the PR to:
1. Rename the fixture/test to make this a compatibility case, not standard string length validation.
2. Add `force_exec_validation=True` to the regression test so it verifies the generated model imports/builds successfully.
3. Exclude this fixture from payload validation, with a reason like:
   `compatibility fixture maps string minItems/maxItems to Pydantic length constraints, which is stricter than JSON Schema`

koxudaxi (approved):

PR #3370, opened 2026-06-12
Fix pydantic v2 dataclass lookaround regex engine

DarkaMaul:
A string field whose schema pattern uses a lookaround assertion `(`(?=`, `(?!`, `(?<=`, `(?<!`)`, generated with `--output-model-type pydantic_v2.dataclass`, produced `constr(pattern=...)` without `ConfigDict(regex_engine="python-re")`.

Pydantic v2's default rust regex engine cannot compile lookaround, so building the generated dataclass raises `SchemaError` at class definition time. The `pydantic_v2.BaseModel` output already handles this case.

## Summary by CodeRabbit

* **New Features**
  * Enhanced Pydantic v2 dataclass generation to detect lookaround regex patterns, including when patterns are reachable through referenced aliases, and automatically select the `python-re` regex engine when needed.
* **Bug Fixes**
  * Ensured regex engine configuration is applied early enough so it reliably appears in the generated dataclass output.
* **Tests**
  * Added/updated OpenAPI and JSON Schema coverage and expected fixtures for lookaround-constrained dataclasses, including reference scenarios, and adjusted relevant conformance exclusions.


src/datamodel_code_generator/model/pydantic_v2/dataclass.py
```diff
@@ -113,6 +116,9 @@ def __init__(  # noqa: PLR0913
             config_attributes_v2_11=self._CONFIG_ATTRIBUTES_V2_11,
         )
 
+        if has_lookaround_pattern(self.fields):
```
koxudaxi: This only catches lookaround patterns attached directly to this dataclass field tree. JSON Schema still generates several lookaround cases through type aliases, e.g. `type LookaroundNullableString = Annotated[constr(pattern=...), Field(...)]` and the dataclass field is only `value: LookaroundNullableString | None`. In that shape `has_lookaround_pattern(self.fields)` does not see the `constr(pattern=...)`, so the generated dataclass still lacks `ConfigDict(regex_engine="python-re")` and fails at import time with Pydantic's rust regex engine.

Please extend the detection to patterns reachable through generated aliases/root types as well, and add/remove coverage for the existing dataclass lookaround exclusions such as `jsonschema/lookaround_anyof_nullable.json`, `lookaround_dict.json`, `lookaround_union_types.json`, and `nested_lookaround_array.json`.

koxudaxi (changes requested):
Thank you for creating this PR. I left one review comment.

src/datamodel_code_generator/model/pydantic_v2/dataclass.py
DarkaMaul: Good catch - updated!

koxudaxi (approved):

PR #3391, opened 2026-06-13
feat: auto-disable black/isort with UserWarning when not installed

mvanhorn:
## Summary

Auto-disables `black` and `isort` formatters with a `UserWarning` when they are not installed, falling back to the built-in formatter instead of crashing with `ImportError`.

## Why this matters

Closes #3058. Users who consume `datamodel-code-generator` via the Python API (`generate_dynamic_models`) in lightweight environments should not be forced to install `black` and `isort`. Today, the default formatter list includes both, causing a hard `ImportError` if either is absent. With this change, missing formatters are detected eagerly in `CodeFormatter.__init__`, a clear `UserWarning` is emitted, and the built-in formatter activates automatically - zero behaviour change for installs that have `black`/`isort` present.

## Changes

- `src/datamodel_code_generator/format.py`: Adds `_is_formatter_available(formatter)` helper that checks importability without raising. In `CodeFormatter.__init__`, calls this helper for `Formatter.BLACK` and `Formatter.ISORT` before attempting to load them; emits `UserWarning` and drops the unavailable formatter from the active list. If no external formatters remain, `Formatter.BUILTIN` is automatically added.
- `tests/test_format.py`: Six new parametric tests covering black-missing, isort-missing, both-missing (BUILTIN fallback), both-present (no warning), explicit-BUILTIN (no warning), and black-only-missing-with-isort-still-active scenarios.

Fixes #3058

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling when requested external formatters (black/isort) are unavailable: the formatter now warns clearly and fails with an actionable error message, suggesting installation or switching to builtin formatters via `--formatters builtin` (no silent fallback when externals are missing).

* **Tests**
  * Updated/added coverage for missing black/isort scenarios, ensuring correct warnings and error messaging, and verifying builtin selection bypasses external checks.

### DIFF
--- a/src/datamodel_code_generator/format.py
+++ b/src/datamodel_code_generator/format.py
@@ -186,6 +186,21 @@ def _get_isort() -> Any:
     return _isort
 
 
+def _is_formatter_available(formatter: Formatter) -> bool:
+    """Return True if the external package required by *formatter* can be imported."""
+    if formatter.value == "black":
+        try:
+            import black  # noqa: F401, PLC0415
+        except ImportError:
+            return False
+    elif formatter.value == "isort":
+        try:
+            import isort  # noqa: F401, PLC0415
+        except ImportError:
+            return False
+    return True
+
+
 class DatetimeClassType(Enum):
     """Output datetime class type options."""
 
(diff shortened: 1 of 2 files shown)


koxudaxi:
@mvanhorn 
Thanks for the PR. In the future, I’d like to make the built-in formatter the default.
That said, I’m not sure we need automatic fallback here. If users want to avoid black/isort, they can explicitly use --formatters builtin. For now, maybe it would be enough to improve the warning/error message and suggest using --formatters builtin instead.

mvanhorn:
Reworked per your direction in d9f9bc1: the automatic fallback is gone. When black or isort is requested but not importable, the CodeFormatter now raises a clear ImportError that names the missing formatter and suggests installing it or passing --formatters builtin. Tests updated accordingly (106 passing locally, ruff clean). This also keeps the door open for making builtin the default later without any hidden behavior switch.

PR #3401, opened 2026-06-14
Reuse isort formatter config

koxudaxi:


koxudaxi:
Closing because this cache approach duplicates isort's config discovery, adds cold-path overhead, and can change settings_path behavior for symlinked paths.

PR #3412, opened 2026-06-14
Fast path list output

koxudaxi:
## Summary by CodeRabbit

* **New Features**
  * `--list-deprecations` and `--list-experimental` commands now support multiple output formats (table, json, markdown)
  * Commands accept flexible argument syntax in both space-separated and equals notation


koxudaxi:
Closing because this optimization is too narrow for the current performance work.

PR #3413, opened 2026-06-14
Reuse JSON Schema ref facts

koxudaxi:
## Summary by CodeRabbit

* **Refactor**
  * Improved JSON Schema reference resolution performance by introducing typed caching for schema data type facts, reducing redundant validation work during parsing, especially for definitions with import metadata.

* **Tests**
  * Added comprehensive test coverage for schema reference caching behavior, verifying both internal and external definition references.


koxudaxi:
Closing because pre-seeding JSON Schema ref facts is too risky for the current optimization pass. It can bypass _load_ref_schema_object() paths where /anchor/current-root/external-ref side effects and warning order are hard to prove equivalent, while the expected gain is small.

PR #3420, opened 2026-06-16
Guard JSON-pointer list indexing against invalid segments

DarkaMaul:
`get_model_by_path` and the `_split_json_pointer` slow path indexed list nodes via a bare `schema[int(key)]`.

A $ref segment hitting an array could crash (ValueError/IndexError) or silently pick the wrong element. Now validated against the RFC 6901 array-index grammar and bounds-checked, raising a clean Error; valid pointers unchanged.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced JSON Pointer array index handling to comply with RFC 6901 standard, providing stricter validation and proper error handling for invalid indices, including non-decimal formats, negative values, and out-of-range references.

* **Tests**
  * Expanded test coverage for array index path segments and validation of invalid JSON Pointer list indices.

### DIFF
--- a/src/datamodel_code_generator/parser/jsonschema.py
+++ b/src/datamodel_code_generator/parser/jsonschema.py
@@ -106,6 +106,22 @@ def unescape_json_pointer_segment(segment: str) -> str:
     return unquote(segment.replace("~1", "/").replace("~0", "~"))
 
 
+_JSON_POINTER_ARRAY_INDEX = re.compile(r"0|[1-9][0-9]*")
+
+
+def _resolve_json_pointer_array_index(sequence: list[YamlValue], segment: object) -> YamlValue:
+    """Resolve a JSON-pointer segment against a list per the RFC 6901 array-index grammar."""
+    text = str(segment)
+    if not _JSON_POINTER_ARRAY_INDEX.fullmatch(text):
+        msg = f"Invalid JSON pointer array index {text!r}: expected a non-negative integer."
+        raise Error(msg)
+    index = int(text)
+    if index >= len(sequence):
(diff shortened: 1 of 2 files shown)


src/datamodel_code_generator/parser/jsonschema.py
```diff
@@ -106,6 +106,22 @@ def unescape_json_pointer_segment(segment: str) -> str:
...
+_JSON_POINTER_ARRAY_INDEX = re.compile(r"0|[1-9][0-9]*")
+
+
+def _resolve_json_pointer_array_index(sequence: list[YamlValue], segment: object) -> YamlValue:
+    """Resolve a JSON-pointer segment against a list per the RFC 6901 array-index grammar."""
+    text = str(segment)
+    if not _JSON_POINTER_ARRAY_INDEX.fullmatch(text):
+        msg = f"Invalid JSON pointer array index {text!r}: expected a non-negative integer."
+        raise Error(msg)
+    index = int(text)
```
koxudaxi: This can still raise ValueError for very long digit strings on Python 3.11 and newer because of sys.get_int_max_str_digits(). Can we wrap int(text) so this path still raises Error?

koxudaxi (changes requested):
Thank you for creating the PR. I left a small comment. Could you take a look?

src/datamodel_code_generator/parser/jsonschema.py
DarkaMaul: Dammit. I tried to be clever here but forgot this parameter. 
Let me update it !

DarkaMaul: Updated!

koxudaxi (approved):
LGTM

PR #3453, opened 2026-06-20
Sync RootModel config template data

koxudaxi:


koxudaxi:
Closing this because the added coverage only exercised RootModel internals. The public generation paths for RootModel regex_engine already pass, so this should not stay as a PR without an end-to-end reproduction.

PR #3489, opened 2026-06-24
Cache structured import requirements

koxudaxi:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved type import generation so nullable/union cases stay consistent even when import results are reused from cache.
  * Fixed JSON Schema “schema-affecting” keyword detection so it’s correctly scoped to each schema object shape.
* **Tests**
  * Expanded coverage for cache normalization and bounded cache eviction behavior.
  * Added a unit test ensuring JSON Schema raw-key caching is memoized per schema object type.


gaborbernat (approved):
LGTM

PR #3552, opened 2026-07-02
Cache builtin formatter source lines

koxudaxi:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved source-segment extraction to better handle UTF-8 and CRLF/newline offset scenarios.
  * Added a safer fallback when complete AST location information isn’t available.

* **Tests**
  * Expanded formatting tests to verify correct extraction for non-ASCII/UTF-8 content.
  * Added checks that repeated extractions reuse cached split source lines and return `None` when location metadata is incomplete.

### DIFF
--- a/src/datamodel_code_generator/_builtin_formatter.py
+++ b/src/datamodel_code_generator/_builtin_formatter.py
@@ -9,7 +9,7 @@ import tokenize
 from collections import defaultdict
 from enum import IntEnum
 from io import StringIO
-from typing import TYPE_CHECKING, Any, TypeGuard
+from typing import TYPE_CHECKING, Any, TypeGuard, cast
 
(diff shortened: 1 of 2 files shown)


koxudaxi:
Local benchmark check:

- Synthetic 240KB generated-code-like input, `apply_builtin_formatter`: main median 3.8560s, this PR median 0.1836s after the typing fix (~21x faster).
- `tests/data/performance/openapi_large.yaml` end-to-end with builtin formatter: main median 0.2666s, this PR median 0.2662s.
- Output SHA256 matched for both benchmark inputs.

Benchmarks were run against temporary worktrees for `origin/main` and `origin/formatter/source-segment-cache`.

PR #3600, opened 2026-07-11
Retain model import caches

koxudaxi:
## Summary by CodeRabbit

* **Performance**
  * Retain reusable import caches for built-in model generation to reduce redundant work.
  * Restrict cache clearing to cases where it could affect customized or changed types.

* **Bug Fixes**
  * Improved correctness for aliasing, generics, discriminator handling, defaults, and cross-module references.
  * Ensured external cache hooks are preserved across module boundaries and rechecked when inherited enum types replace built-in wrappers.

* **Tests**
  * Added integration coverage for cache retention vs invalidation, alias/generic scenarios, cross-module hook behavior, and inherited enum rechecking.

### DIFF
--- a/src/datamodel_code_generator/parser/base.py
+++ b/src/datamodel_code_generator/parser/base.py
@@ -2249,8 +2249,9 @@ class Parser(ABC, Generic[ParserConfigT, SchemaFeaturesT]):
                         try:
                             converted_default = set(model_field.default)
                         except TypeError:
-                            # Elements are not hashable (e.g., contains dicts)
-                            # Skip both type and default conversion to keep consistency
+                            # Keep the unhashable default unchanged. Nested list types may
+                            # already have been converted in place, so invalidate their imports.
+                            _clear_model_imports_cache_if_retained(model, can_retain_cache=can_retain_cache)
                             continue
                         model_field.default = converted_default
                     self.generation_store.replace_field_type(model_field, set_data_type)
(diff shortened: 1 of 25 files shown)


koxudaxi:
Addressed the review against the current code in `53e74345`:

- Unique-list cache invalidation: the normal replacement path already clears the owner cache through `GenerationStore.replace_field_type()`. I found and fixed the real gap: a nested list can be mutated in place before an unhashable default raises `TypeError`, which skipped `GenerationStore`. The added external schema/golden E2E combines duplicate inline models, `reuse_model`, `collapse_reuse_models`, a warmed imports cache, and the unique-list mutation. It asserts `Item:invalidated`, while unchanged `Left` and `Model` remain retained. Generated output matches `origin/main`.
- Final `validate_default` ordering: left unchanged after verification. `origin/main` uses the same order; the first pass sets the required flag before the imports snapshot, and the built-in final pass is idempotent. Moving the pass changes observable timing for external custom fields/hooks.
- Generic-base fallback: left unchanged after verification. The non-retain path already performs the legacy blanket clear before generic-base processing, and `GenerationStore.set_base_classes()` clears on the following mutation. An extra clear changes the externally observable custom hook history and generated golden output.
- Mutation-ordering coverage: the new compound E2E covers the reported unique/reuse/collapse gap. Existing required-field override tests cover the `GenerationStore` insert/remove invalidation paths. `get_dedup_key()` invalidation beyond imports is pre-existing on `main` and is not changed by this PR.

Validation: `5518 passed, 16 skipped`; 1,827 generated files compiled and 79 executed; Ruff, ty, changed-file hooks, GenerationStore guard, and diff checks pass.

PR #3622, opened 2026-07-23
Invalidate generation facts for every list mutation

koxudaxi:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when modifying generated model lists, including sorting, reversing, extending, and in-place updates.
  * Ensured derived generation information is refreshed even when a list operation encounters an error or partially completes.

* **Tests**
  * Expanded coverage for all supported list mutation behaviors, return values, and error scenarios.
  * Added checks to ensure future list operations maintain the required invalidation behavior.


koxudaxi:
Addressed the CodeRabbit suggestion in c120a01a with a sort key that succeeds once and then raises, proving the `finally` path invalidates generation facts after a partial sort attempt.

PR #3667, opened 2026-07-30
Preserve type overrides during model reuse

koxudaxi:
## Summary by CodeRabbit

* **Bug Fixes**
  * Ensured model reuse and deduplication respect `--type-overrides`, so overridden model types remain unchanged across tree and module reuse.
  * Kept generated output consistent when `--collapse-reuse-models` is enabled.
* **Tests**
  * Added end-to-end coverage verifying type overrides are preserved for both standard and collapsed reuse modes.
  * Added execution-time validation to confirm generated models are correct.


abdullahra634-dotcom:
@koxudaxi @kriberg @yarikoptic @natikgadzhi

PR #3704, opened 2026-08-05
Speed up parser hot paths

koxudaxi:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of field-name collisions during model generation.
  * Corrected detection of `readOnly` and `writeOnly` properties across composed JSON Schemas.
  * Generated models now correctly represent read-only, write-only, and shared fields in composed schemas.

* **Tests**
  * Added coverage for generating Pydantic models from JSON Schemas with composed read/write properties.


koxudaxi:
@coderabbitai review

PR #3871, opened 2026-09-05
Test dynamic model cache ordering

koxudaxi:
## Summary by CodeRabbit

* **Tests**
  * Expanded coverage for schema ordering across fields, nested fields, definitions, and OpenAPI components.
  * Added checks confirming input data remains unchanged during model generation.
  * Added validation for cache hits and misses, model reuse, field ordering, and JSON output consistency.

### DIFF
--- a/tests/data/dynamic_models/cache_defs_model_order_first.json
+++ b/tests/data/dynamic_models/cache_defs_model_order_first.json
@@ -1,0 +1,44 @@
+{
+  "title": "DefinitionsEnvelope",
+  "$defs": {
+    "Alpha": {
+      "type": "object",
+      "properties": {
+        "first": {
+          "type": "string"
+        },
+        "second": {
+          "type": "integer"
+        }
+      },
+      "required": [
+        "first",
+        "second"
+      ]
+    },
+    "Beta": {
+      "type": "object",
+      "properties": {
+        "third": {
+          "type": "boolean"
+        },
+        "fourth": {
+          "type": "number"
+        }
+      },
+      "required": [
+        "third",
+        "fourth"
+      ]
+    }
+  },
+  "type": "object",
+  "properties": {
+    "context": {
+      "type": "string"
+    }
+  },
+  "required": [
+    "context"
+  ]
+}
--- a/tests/data/dynamic_models/cache_defs_model_order_second.json
+++ b/tests/data/dynamic_models/cache_defs_model_order_second.json
@@ -1,0 +1,44 @@
+{
+  "title": "DefinitionsEnvelope",
+  "$defs": {
+    "Beta": {
+      "type": "object",
+      "properties": {
+        "third": {
+          "type": "boolean"
+        },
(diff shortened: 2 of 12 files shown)


koxudaxi:
The production cache-key fix was already merged in #3865. This PR adds focused regression coverage for that behavior.

Schemas with equal values but different insertion order must produce the same observable model/field order and validated JSON as `cache_size=0`. Such inputs intentionally use distinct cache entries and classes; repeating the same ordered schema and configuration still reuses the cached dictionary and classes.

The four fixture families isolate root properties, nested properties inside array items, `$defs` order, and OpenAPI component order. Each runs in both call orders with a real `MappingProxyType` input and real Pydantic validation. Existing assertion helpers compare external fixtures and guard input values; serialized snapshots also detect input key reordering.

Local validation:

- Restoring only `sort_keys=True` in the current source makes all four ordering cases fail.
- Dynamic-model and assertion-helper tests: **86 passed, 1 skipped**; `dynamic.py` and `test_dynamic_models.py` have **100% line and branch coverage** (639 statements, 108 branches).
- Standard locked suite: **8,418 passed, 16 skipped**, using `uvx --python 3.13 --with tox-uv tox run -e py314-nocov-parallel -- tests -n 4 --dist=worksteal -m 'not perf' --tb=short -q --no-cov`.
- Standard `tox run -e type` and the configured pre-commit checks pass.
- The existing cache tests exercise configuration/module separation, target filtering, disabled caching, serialization fallback, bounded FIFO eviction, clear, concurrency, and compact digest keys.

Performance verification compares the same `bde9cc71` source with only the schema `sort_keys=True` expression restored for the before variant. Both use CPython 3.14.0, Pydantic 2.12.5, identical dependencies, the built-in formatter, and timestamps disabled on an Apple M4 Max. Five alternating subprocess rounds; timing and tracemalloc are separate.

| Input | Cache mode | Before median ms | After median ms |
| --- | --- | ---: | ---: |
| Small (1 class) | Hit | 0.008854 | 0.008584 |
| Small | Miss | 1.079333 | 1.086208 |
| Small | Disabled | 1.125250 | 1.067709 |
| Large (501 classes) | Hit | 1.101958 | 0.796623 |
| Large | Miss | 325.821583 | 326.302416 |
| Large | Disabled | 328.055083 | 324.796416 |

The large-hit improvement is 27.71%; traced peak allocation falls from 364.29 to 331.59 KiB. Miss timings have overlapping round intervals. The key remains 139 characters (180 bytes for the string object) for a 169,191-byte schema. Repeated-hit live memory plateaus at the same values before/after; no additional retention was observed in these bounded runs. These local measurements do not establish performance for every schema, platform, or allocator.

An exploratory `--all-extras` environment selected the separate `ryaml` backend and failed one existing PyYAML-specific test on both baseline and branch. The standard locked `extras=all` tox environment excludes that separate extra and passes the full suite above; no unrelated expectation was changed.

PR #3880, opened 2026-09-05
Fix relative YAML cache paths

koxudaxi:
## Summary by CodeRabbit

- **Documentation**
  - Clarified OpenAPI scope behavior for parameter models, request bodies, responses, and referenced parameters.
  - Updated guidance and examples for including URL path parameters in generated models.

- **Bug Fixes**
  - Improved YAML schema caching so relative and absolute paths are handled consistently, preventing distinct schemas from being confused.
  - Preserved correct behavior when loading schemas from different directories and through symlinks.


koxudaxi:
The CodSpeed report identifies different runtime environments and reports similar 30–40% changes across many unrelated generators and runtime helpers. I checked this change locally using alternating isolated baseline/patched processes with the other test workers paused.

Absolute cached YAML loads measured 1.512 µs → 1.581 µs. Relative loads measured 1.340 µs → 8.566 µs because the cache key must obtain the caller's current directory; relative loading followed by real API generation measured 493.432 µs → 504.082 µs. Every normal-output hash matched, and repeated loads retained cached object identity. No reparsing, deep copy, directory scan, or symlink resolution was introduced. The absolute path avoids the cwd lookup.

The correctness workflows passed and Codecov reports 100% project and patch coverage. The hosted-runner CodSpeed comparison is retained as a separate noisy result; it was not used to justify changes to normal generated output.

PR #3883, opened 2026-09-05
Fix dataclass Field imports

koxudaxi:
## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Pydantic v2 dataclass generation for fields with dictionary defaults.
  - Dictionary defaults now generate appropriate `Field` declarations, preserving independent mutable defaults.
  - Prevented unnecessary `Field` imports or declarations for class variables and required dictionary fields without defaults.

- **Tests**
  - Added coverage for API and CLI generation, type aliases, field constraints, mutable-default isolation, and related edge cases.


koxudaxi:
CodSpeed reports different runtime environments and a broad 32% slowdown, including unrelated msgspec and TypedDict benchmarks. The isolated alternating comparison for this change measured API 0.861 to 0.877 ms with overlapping ranges, cold CLI 112.684 to 113.186 ms, and identical 195,743-byte traced peaks. Normal output hashes match; the bug fixture differs only by the missing Field import. Astra independently re-reviewed the exact 8f1655d head, reran 28 related tests plus the forced-required control, and verified complete coverage of the modified predicate. All correctness and coverage checks pass; the remaining CodSpeed result is consistent with runner noise.

PR #4073, opened 2026-09-15
Fix builtin formatter wrapping of boolean expressions

koxudaxi:
## Summary by CodeRabbit

- **New Features**
  - The built-in formatter now supports boolean expressions in synchronous and asynchronous functions and methods.
  - Nested boolean groups wrap recursively when needed while preserving conditional prefixes, trailing guards, comments, docstrings, assignment expressions, and Unicode names.
  - Long asynchronous function definitions and same-line return annotations are formatted consistently.

- **Bug Fixes**
  - Avoided unnecessary multiline wrapping when expressions fit the configured line length.
  - Prevented duplicate processing of boolean-expression candidates.
  - Leading blank lines are normalized before formatting.


koxudaxi:
Addressed async-signature support and the return-annotation follow-up in e4831e6b. The renderer now finds the parameter-list boundary before the return annotation using its AST UTF-8 offset and preserves the annotation verbatim. Multiline annotations remain unchanged. The original Annotated[str, Field(...)] example retained a valid AST, but Literal[")"], quoted annotations, and triple-quoted annotations exposed a real async regression; those are fixed.

The existing CLI/API E2E fixture now executes an async method with a Literal annotation and a synchronous method with Unicode parameters and a quoted annotation. The Unicode signature is 111 characters long, so width 88 exercises the byte-offset split. Widths 88 and 120 also verify full generated output and formatting idempotence. All 12 formatter E2E cases pass. Fable and Astra Max reviewed the fix; Astra Max checked 84 syntax/AST/idempotence cases, including multiline and triple-quoted annotations.

The conjunction in builtin_boolean_wrapping/and.py is intentional: this custom formatter fixture exercises `and` with JSON dictionary inputs, and its paired `or` fixture exercises disjunction. Production schema-validation helpers use `or`; #4074 separately checks real dict, UserDict, and MappingProxyType payloads.

The compound-guard and trailing-condition fixes preserve the surrounding condition. The dedicated validator formatter fixtures use shared Black/builtin goldens. More complex pre-existing layouts remain outside this formatter's supported Black subset; local review verified unchanged ASTs and idempotence for the reported cases.

Final stack validation at 09716baa: 18,946 tests pass (15 skipped) in each of the builtin-default and Black-parity configurations. The final 50 related E2E cases cover all 145 changed executable lines. Both PRs have 67 successful checks, one skipped check, and only the permitted CodSpeed analysis failure. The additional `multiple_groups_guard` fixture comment is answered [here](https://github.com/datamodel-code-generator/datamodel-code-generator/pull/4073#issuecomment-5684654292) with its source AST and explicit runtime expectations.

koxudaxi:
Final local benchmark: CPython 3.12.9, main 3452dfbb versus stack head 09716baa (this formatter change plus #4074). Each result is the median of 9 repeats; small inputs use 30 generations per repeat and large inputs use 1. Tests were stopped during measurement. The second run reversed baseline/head order.

| Generation with builtin formatter | Main (ms) | Stack (ms) | Main, reverse order (ms) | Stack, reverse order (ms) |
| --- | ---: | ---: | ---: | ---: |
| Small model | 1.271 | 1.297 | 1.280 | 1.249 |
| Large models | 135.085 | 138.506 | 133.741 | 133.506 |
| Frozen large models | 154.597 | 157.353 | 153.322 | 152.580 |
| Ordinary schema validators | 1.601 | 1.694 | 1.600 | 1.672 |
| Long Boolean validator fixture | 1.686 | 1.853 | 1.686 | 1.805 |

Ordinary-generation differences changed sign with measurement order. Validator generation consistently adds about 0.07–0.09 ms (4.5–5.8%), and the long-Boolean fixture adds 0.12–0.17 ms (7.1–9.9%) for formatting the baseline leaves untouched. Peak allocations are unchanged for ordinary generation and the long-Boolean fixture; the standard validator case increases by 848 bytes. These local results are recorded independently of CodSpeed's comparisons across different runtime environments.

koxudaxi:
@coderabbitai review

koxudaxi:
@coderabbitai The `multiple_groups_guard` expression is intentional formatter input. Its [source template](https://github.com/datamodel-code-generator/datamodel-code-generator/blob/e4831e6ba709c09caa83dfcbf5f9495c7e7d12c2/tests/data/templates/builtin_boolean_wrapping/guards/pydantic_v2/BaseModel.jinja2) uses `value is not None or value is False`; the [runtime fixture](https://github.com/datamodel-code-generator/datamodel-code-generator/blob/e4831e6ba709c09caa83dfcbf5f9495c7e7d12c2/tests/data/expected/main/jsonschema/builtin_boolean_wrapping/runtime.txt) explicitly expects `True` for `None` and `False` for `False`. Both generated methods at widths 88 and 120 have the same AST as the input method, and CLI/API E2E executes these payloads. Changing the inner operator to `and` would change the input program's behavior. The two Boolean groups exercise replacement-overlap handling.

Likewise, `builtin_boolean_wrapping/and.py` deliberately covers conjunction formatting. Production Mapping validation and real non-dict Mapping payloads are covered separately in #4074; the fixture's intent and the resolved async-annotation issue are detailed in [the earlier response](https://github.com/datamodel-code-generator/datamodel-code-generator/pull/4073#issuecomment-5683376005).

PR #4078, opened 2026-09-16
Fix release draft inline code validation and use Opus 5

koxudaxi:
## Summary by CodeRabbit

- **Bug Fixes**
  - Release draft validation now correctly permits square brackets and angle brackets within single-backtick inline code.
  - Markdown syntax remains rejected when it appears outside inline code.

- **Chores**
  - Updated automated release analysis to use the latest configured Claude model.

- **Tests**
  - Added end-to-end coverage for accepted inline-code expressions and rejected surrounding Markdown attacks.


koxudaxi:
Validation and security review:

- Reproduced the failure from job 104639686824: the generated release note contains `friends: list[Pet]` inside balanced inline code. That actual output is now an external CLI regression fixture.
- 18 real CLI cases cover safe code, active Markdown before/between/after code, malformed delimiters, Unicode controls, and marker leakage. The changed validator function has 100% line and branch coverage from these CLI cases; the full validator has 100% with its complete test suite.
- GitHub's GFM rendering API confirmed that accepted inline URLs, HTML, and mentions produce code elements, without links or HTML elements. Read-only tool permissions, marker checks, strict reasoning schema, and shell-safe output handling remain in place.
- Local benchmark, median of seven runs of 100,000 calls: plain text 1.841 → 1.825 µs, simple inline code 1.425 → 1.427 µs; newly accepted generic types take 1.060 µs. The former implementation rejected those types.
- The CodSpeed report flags different runtime environments and unstable hosted-runner walltime. This PR does not change generator runtime code, so its reported generator regressions are not evidence of a regression in the modified validator.
- Release Draft runs only after merge. Opus 5's actual connection using the existing OAuth token remains unverified before merge.

koxudaxi:
Post-merge verification: Release Draft run 35049049091 completed successfully. The execution log confirms `claude-opus-5`, and both `analyze` (including output validation) and `update-draft` passed using the existing OAuth authentication.

https://github.com/datamodel-code-generator/datamodel-code-generator/actions/runs/35049049091

